PR #1, opened 2018-08-04
Use an array of structures instead of a structure of arrays for histograms

ogrisel:
On master (structure of arrays using `@njitclass`):

```
$ python benchmarks/bench_histogram.py 
Building feature histogram on 10000000 values of 100000000
Built in 2.007s
```

On this branch / PR (array of structures):

```
$ python benchmarks/bench_histogram.py 
Building feature histogram on 10000000 values of 100000000
Built in 1.655s
```

So the memory access pattern is probably better this way.

Questions for @seibert:

- is there a natural way to use a @jitclass to implement the array of structures pattern.
- when using the `@jitclass` construct, it does not seem to be possible to use `introspect_types` / asm on the instance methods. Is this is a missing feature in the current implementation or is there a fundamental problem that prevents implementing those?

### DIFF
--- a/benchmarks/bench_histogram.py
+++ b/benchmarks/bench_histogram.py
@@ -17,8 +17,11 @@ def make_data(n_bins=256, n_samples=int(1e8), n_subsample=int(1e6),
     binned_feature = rng.randint(0, n_bins - 1, size=n_samples)
     binned_feature = binned_feature.astype(np.uint8)
 
-    sample_indices = rng.choice(np.arange(n_samples, dtype=np.uint32),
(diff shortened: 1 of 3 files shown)


ogrisel:
I have just pushed a new commit with unrolled histogram building as done in LightGBM. Here are the results:

```
$ python benchmarks/bench_histogram.py 
build_histogram on 1e+07 values of 1e+08
Built in 1.771s
build_histogram_unrolled on 1e+07 values of 1e+08
Built in 0.863s
```

PR #3, opened 2018-08-06
naive implementation of exact greedy gbt

lancifollia:
@ogrisel 
This is a code dump of [tinygbt](https://github.com/lancifollia/tinygbt).

Also, I added a simple benchmark code & data.

- `/benchmark/bench_exact_greedy.py`
- `/benchmark/data/regression.(test|train)`

This data is copy of [LightGBM's regression example data](https://github.com/Microsoft/LightGBM/tree/master/examples/regression).

You can test this with following command.

```
$git clone https://github.com/lancifollia/pygbm.git pygbm_exact_greedy
$cd pygbm_exact_greedy
$git checkout exact_greedy
$cd benchmarks
$python bench_exact_greedy.py
```


ogrisel:
Thanks. I am currently making some progress on an alternative tree grower implementation with a slightly different API based on histogram splitting for binned numerical features.

I will review your PR in light of that work (probably in a couple of days). I think the main Gradient Boosting loop of your code can be reused almost directly. The exact tree learning will probably need a refactor to match my implementation though.

I'll ping you when I am ready.

ogrisel:
@lancifollia sorry for the lack of feedback on this PR. @NicolasHug and I have been working on making the binned feature histogram case as efficient as possible and we want to optimize the global design of the library to suite to make sure that we can get the maximum performance on with binned feature histogram first. Once done we will probably refactor the grower.py / splitting.py modules a bit to allow for exact splits as an option.

As you can see in master we made good progress and running pygbm on the Higgs boson benchmark from the `benchmarks/` subfolder yields performance competitive with that of LightGBM.

@NicolasHug will work on adding the missing features such as: adding a proper classification loss, dealing with categorical variables, sparse data and missing values  and so on.

Feel free to chip in to review the ongoing pull request and let us know if there is something else you would like to improve in the mean time.

PR #12, opened 2018-09-19
Added plotting tool to compare trees with LightGBM

NicolasHug:
Basic plotting tool (requires graphviz) that may make debugging easier. I doubt this is going to stick so I didn't add tests but I can if needed.

example:

out


ogrisel:
Thanks. At some point we should add a test, but it's fine for now.

PR #13, opened 2018-09-21
[MRG] Add support for classification tasks

NicolasHug:
### Udpate

This PR:

- Introduces `GradientBoostingRegressor` and `GradientBoostingClassifier` classes, with `BaseGradientBoostingMachine` as the base class
- Introduces loss classes `LeastSquares` and `Logistic`
- slightly updates the leaf value computation to be consistent with LightGBM (just multiplicative factor of `-1`), and updates tests consequently
- Adds some tests for `GradientBoostingClassifier` and the loss functions
- updates the higgs boson benchmark to use classifiers

--------

### Outadted

- Created `pygbm/loss.py` with least squares loss
- Updated leaf value computation so that it fits with that of XGBoost paper (not set in stone)
- Updated tests consequently

One consequence of the new implementation is that the `gradients` array has to be recomputed (reallocated) at every iteration now. We cannot just simply update it like before (`gradients += prediction_of_current_tree`), because I don't think this would work for other losses.

LMK if this is going in the right direction and I'll try adding a classification loss and split GradientBoostingMachine into 2 subclasses.

:)


pygbm/gradient_boosting.py
```diff
@@ -100,7 +103,9 @@ def fit(self, X, y):
                 bin_thresholds=self.bin_mapper_.bin_thresholds_)
             predictors.append(predictor)
             self.n_iter_ += 1
-            gradients -= predictor.predict_binned(X_binned_train)
+            y_pred = self._predict_binned(X_binned)
```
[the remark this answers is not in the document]
ogrisel: Maybe do:

```python
self.n_iter_ = 0
y_pred = np.zeros(y)
gradients = self._loss.get_gradients(y, y_pred)
hessians = self._loss.get_hessians(y, y_pred)
while True:
    ...
    y_pred += self._predict_binned(X_binned)
    gradients = self._loss.get_gradients(y, y_pred)
    hessians = self._loss.get_hessians(y, y_pred)
```

pygbm/loss.py
```diff
@@ -0,0 +1,16 @@
+import numpy as np
+
+
+# TODO: ABC for losses, regression losses and classification losses
+class LeastSquares:
+    def __init__(self):
+        self.hessians = np.full(shape=1, fill_value=2, dtype=np.float32)
+
+    def get_gradients(self, y, y_pred):
+        return -2 * (y - y_pred).astype(np.float32)
```
ogrisel: Cosmetics:

```python
return 2 * (y_pred - y).astype(np.float32)
```

ogrisel: Maybe the `y` variable in the loss functions could be renamed to `y_true` (but not in the fit method where `y_train` / `y_val` is better.

tests/test_predictor.py
```diff
@@ -17,8 +18,9 @@ def test_boston_dataset():
     X_train_binned = mapper.fit_transform(X_train)
     X_test_binned = mapper.transform(X_test)
 
-    gradients = y_train.astype(np.float32)
-    hessians = np.ones(1, dtype=np.float32)
+    loss = LeastSquares()
+    gradients = loss.get_gradients(y_train, 0)
+    hessians = loss.get_hessians(y_train, 0)
```
ogrisel: This is not related to the predictor module. Please create a new test file named `test_losses.py` dedicated to unit test for the loss classes and then another test file named `test_gradient_boosting.py` that runs high level test for the estimator public API.

ogrisel:
Also please add the deviance / logistic loss for classification to this PR with tests to ensure that the design makes sense.

pygbm/loss.py
```diff
@@ -0,0 +1,16 @@
+import numpy as np
+
+
+# TODO: ABC for losses, regression losses and classification losses
+class LeastSquares:
+    def __init__(self):
+        self.hessians = np.full(shape=1, fill_value=2, dtype=np.float32)
```
ogrisel: Please make that attribute private: `self._constant_hessian = np.full(shape=1, fill_value=2, dtype=np.float32)`.

pygbm/loss.py
```diff
@@ -0,0 +1,16 @@
...
+
```
ogrisel: Please add a `__call__` method that computes the actual loss value and use it in tests to make sure that the derivatives agree with the loss: e.g. that a minimum is reached when taking 1d newton steps from some arbitrary init: try with `y_true = [42.]`, `y_pred_init = [-12]` and use:

https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.newton.html

to iteratively update `y_pred` until reaching convergence and check that the loss value at convergence is 0 (e.g. for least squares regression and for classification with the binomial deviance, that is with `y_true` with values in `{0, 1}`).

ogrisel: You can also try write tests with 5D `y_true` / `y_pred` arrays: small dimensions that makes it easy to write test with hardcoded values in the test but not as trivial as the 1D case.

pygbm/loss.py
```diff
@@ -0,0 +1,16 @@
+import numpy as np
+
+
+# TODO: ABC for losses, regression losses and classification losses
```
ogrisel: I am not sure we actually need an ABC in this case. We can just use duck typing. Let's only use a base class if there is some redundant code to factorize across loss functions.

tests/test_predictor.py
NicolasHug: I wasn't trying to test the loss here, I just had to change the hardcoded gradients / hessians computation for the test to pass.

This test implicitly used a least squares loss, and as I changed the way the leaf values are computed, I also had to change the gradient / hessians computation for the test to pass. I can change it to 

```
    gradients = -2 * y_train
    hessians = np.full(shape=1, fill_value=2)
```
but that's still implicitely using least squares.

Of course I will subsequently add `test_losses.py` and `test_gradient_boosting`.

ogrisel: Alright.

pygbm/grower.py
```diff
@@ -159,7 +159,7 @@ def _finalize_leaf(self, node):
         XGBoost: A Scalable Tree Boosting System, T. Chen, C. Guestrin, 2016
         https://arxiv.org/abs/1603.02754
         """
-        node.value = self.shrinkage * node.sum_gradients / (
+        node.value = -self.shrinkage * node.sum_gradients / (
             node.sum_hessians + self.splitter.l2_regularization)
```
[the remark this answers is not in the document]
ogrisel: Actually, I am not so sure anymore.

It also more intuitive to have the prediction of the ensemble to be directly the sum of all the individual predictors without a leading minus sign.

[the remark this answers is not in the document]
ogrisel: Maybe we should have a look at how the predict and predict_probas function are implemented in LightGBM.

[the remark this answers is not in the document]
NicolasHug: Looks like LGBM also uses the minus sign [for the leaves values](https://github.com/Microsoft/LightGBM/blob/master/src/treelearner/feature_histogram.hpp#L502).

`output` is equal to `sum_g / (sum_h + l2)` so the result is indeed `sum_g^2 / (sum_h + l2)`, ignoring the l1 regularization which is 0 by default.

tests/test_compare_lightgbm.py
```diff
@@ -51,21 +53,99 @@ def test_same_predictions_easy_target(seed, min_samples_leaf, n_samples,
...
+                                          max_leaf_nodes=max_leaf_nodes)
 
     est_lightgbm.fit(X_train, y_train)
     est_pygbm.fit(X_train, y_train)
 
     pred_lgbm = est_lightgbm.predict(X_train)
     pred_pygbm = est_pygbm.predict(X_train)
-    np.testing.assert_array_almost_equal(pred_lgbm, pred_pygbm, decimal=3)
+    # less than 1% of the predictions are different up to the 3rd decimal
+    assert np.mean(abs(pred_lgbm - pred_pygbm) > 1e-3) < .01
```
NicolasHug: I changed this test because it was failing for a  **single** sample with `seed=4`. I think this is because I changed the dataset generation to something not so easy but I guess it's better this way? Can revert if needed.

tests/test_compare_lightgbm.py
```diff
@@ -51,21 +53,99 @@ def test_same_predictions_easy_target(seed, min_samples_leaf, n_samples,
...
 
-    if max_leaf_nodes < 10 and n_samples > 1000:
+    if max_leaf_nodes < 10 and n_samples >= 1000:
         pred_lgbm = est_lightgbm.predict(X_test)
         pred_pygbm = est_pygbm.predict(X_test)
-        np.testing.assert_array_almost_equal(pred_lgbm, pred_pygbm, decimal=3)
+        # less than 1% of the predictions are different up to the 4th decimal
+        assert np.mean(abs(pred_lgbm - pred_pygbm) > 1e-4) < .01
```
NicolasHug: Also changed this one because it was actually not passing even on master (the above condition was never reached as `n_samples<=1000`)

NicolasHug:
FYI now the higgs boson benchmarks gives:

```
python benchmarks/bench_higgs_boson.py --n-trees 500 --learning-rate 0.1 --n-leaf-nodes 255 --subsample 1000000
```


| Model | Time | acc | 
| :--- | ---: | ---: | 
| LightGBM | 73s | 0.7488 
| pygbm | 92s | 0.7423 |

ogrisel:
> FYI now the higgs boson benchmarks gives: ...

Nice! But do you have any idea why the classification loss is so slow in LightGBM? Do you do something special in your implementation?

ogrisel:
Is pygbm always faster? Even with the full dataset on a machine with more than 4 CPU cores?

NicolasHug:
I can only test with my own machine (4 cores) but:

- classification is slower than regression in both libs probably because logistic loss has non-constant hessians
- one reason pygbm is slower than lightgbm is probably because gradients and hessians are uploaded inplace in lightgbm, while they are reallocated at each iteration in pygbm. There is also the fact that in pygbm, hessians in logistic loss need the gradients, so they are computed twice. I made a comment in the code about this, we might want to use a `get_gradients_and_hessians` method to avoid it.


I'll try to re-use the gradients and hessians arrays and compute them with numba and keep you updated about any improvement

NicolasHug:
>But do you have any idea why the classification loss is so slow in LightGBM? Do you do something special in your implementation?

>Is pygbm always faster?

On my machine pygbm is **slower**. Is it faster on yours?

tests/test_compare_lightgbm.py
NicolasHug: In fact, the tests with ``np.testing.assert_array_almost_equal(pred_lgbm, pred_pygbm, decimal=3)`` pass on my machine. They fail on [travis](https://travis-ci.org/ogrisel/pygbm/jobs/459913194) but I can't reproduce the failure, even with the same packages (I'm using the lastest lightgbm version from their master branch so maybe that's why?).

ogrisel:
> On my machine pygbm is slower. Is it faster on yours?

Sorry, I read the other way round. That would have been surprising :)

ogrisel:
> one reason pygbm is slower than lightgbm is probably because gradients and hessians are updated inplace in lightgbm, while they are reallocated at each iteration in pygbm.

I doubt that allocation time in the outer (gradient boosting) loop has any impact on the runtime.

> There is also the fact that in pygbm, hessians in logistic loss need the gradients, so they are computed twice. I made a comment in the code about this, we might want to use a get_gradients_and_hessians method to avoid it.

+1, this is a better design.

ogrisel:
We might also want to have a single method that computes both the objective value, the gradients and the hessians at once. For some loss there might be some computation to mutualise.

NicolasHug:
>I doubt that allocation time in the outer (gradient boosting) loop has any impact on the runtime.

You're probably right. One other major difference that I missed is that gradients and hessians computation is parallelized in LightGBM. That's probably a big low-hanging fruit for us... but as `jitclass` doesn't support `@njit(parallel)` that would make the code a bit ugly (plus, there's no built-in static scheduling).

I'd say we do it anyway? We could keep `get_gradients_slow()` and `get_hessians_slow()` to have a clean version of the methods.

benchmarks/bench_higgs_boson.py
```diff
@@ -69,38 +67,41 @@ def load_data():
...
+    lightgbm_model = LGBMClassifier(objective='binary',
+                                    n_estimators=n_trees,
+                                    num_leaves=n_leaf_nodes,
+                                    learning_rate=lr, verbose=10,
+                                    min_data_in_bin=1)
     lightgbm_model.fit(data_train, target_train)
     toc = time()
     predicted_test = lightgbm_model.predict(data_test)
-    roc_auc = roc_auc_score(target_test, predicted_test)
-    print(f"done in {toc - tic:.3f}s, ROC AUC: {roc_auc:.4f}")
```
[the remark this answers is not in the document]
NicolasHug: I'll put ROC AUC back

I've added `predict_proba` to `GradientBoostingClassifier` already

tests/test_loss.py
```diff
@@ -0,0 +1,48 @@
...
+
+    loss = _LOSSES[loss]()
+    y_true = np.array([42.])
+    y_pred_init = np.array([-12])
+    zero = newton(loss, x0=y_pred_init, args=[y_true])
+
+    assert np.allclose(zero, y_true)
+    assert np.allclose(loss(y_true, zero), 0)
+    assert np.allclose(loss.get_gradients(y_true, zero), 0, atol=1e-7)
+
```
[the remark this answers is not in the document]
NicolasHug: I've tried doing this but can't seem to make it work :/

```python
@pytest.mark.parametrize('loss', ['ls'])
def test_derivatives2(loss):
    # Check that gradients are zero when the loss is minimized on 1D array

    loss = _LOSSES[loss]()
    y_true = np.array([42.])
    y_pred_init = np.array([-12])

    def fprime(y_true, y_pred):
        return loss.get_gradients(y_true, y_pred) * 2

    zero = newton(loss, fprime=fprime, x0=y_pred_init, args=[y_true])

    assert np.allclose(zero, y_true)
    assert np.allclose(loss(y_true, zero), 0)
    assert np.allclose(loss.get_gradients(y_true, zero), 0, atol=1e-7)
```

```
E       RuntimeError: Failed to converge after 50 iterations, value is [-3.4431561e+10]
```

Also note that this test fails for `loss='logistic'`, even with values in [0, 1]. I don't understand why.

ogrisel:
> I'd say we do it anyway? We could keep get_gradients_slow() and get_hessians_slow() to have a clean version of the methods.

At least it worth check whether or not this is causing a significant performance difference.

NicolasHug:
With the parallelized code + reuse of the gradients/hessians arrays the above benchmark runs in 86s (vs 92s) before. I have consistent results over 3 different runs.

pygbm/loss.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def __call__(self, y_true, y_pred, average=True):
+        loss = np.power(y_true - y_pred, 2)
+        return loss.mean() if average else loss
+
+    def update_gradients_and_hessians(self, gradients, hessians, y_true,
+                                      y_pred):
+        return _update_gradients_least_squares(gradients, y_true, y_pred)
+
+    def get_gradients(self, y_true, y_pred):
```
NicolasHug: @ogrisel As we're now using `update_gradients_and_hessians` and as it's inplace, I removed the `get_hessians` methods and only kept `get_gradients` because it's still used in the tests.

Let me know if you think we could get rid of it? Maybe `test_numerical_gradients` is enough?

pygbm/gradient_boosting.py
```diff
@@ -196,8 +214,56 @@ def _should_stop(self, scores):
...
+            min_samples_leaf=min_samples_leaf,
+            l2_regularization=l2_regularization, max_bins=max_bins,
+            max_no_improvement=max_no_improvement,
+            validation_split=validation_split, scoring=scoring, tol=tol,
+            verbose=verbose, random_state=random_state)
+
+    def predict(self, X):
+        return np.argmax(self.predict_proba(X), axis=1)
+
+    def predict_proba(self, X):
```
NicolasHug: This will likely be moved to a method of the `Loss` when we implement multiclass classification, because of the proba is not computed the same for logistic loss and for softmax (at least not when it comes to implementation)

pygbm/loss.py
ogrisel: Maybe it's possible to remove this method from the loss but to rewrite the other test to add an local function in the test that needs `get_gradients`  to actually rely on `update_gradients_and_hessians` under the hood?

ogrisel:
> With the parallelized code + reuse of the gradients/hessians arrays the above benchmark runs in 86s (vs 92s) before. I have consistent results over 3 different runs.

> Not much but definitely worth it IMO.

Still worth it indeed. Too bad that numba jitclass methods do not support `parallel=True`.

NicolasHug:
I removed the `get_gradients()` method but kept the tests intact with a helper.

I'm not sure what to do about the other tests...

pygbm/gradient_boosting.py
```diff
@@ -78,10 +92,10 @@ def fit(self, X, y):
...
         if self.verbose:
             print("Fitting gradient boosted rounds:")
-        # TODO: plug custom loss functions
-        y_pred = np.zeros_like(y_train, dtype=np.float32)
-        gradients = np.asarray(y_train, dtype=np.float32).copy()
-        hessians = np.ones(1, dtype=np.float32)
+        # TODO: is the initial prediction always 0? What about classif?
+        y_pred = np.zeros_like(y_train)
+        gradients, hessians = self.loss_.init_gradients_and_hessians(
+            n_samples=y_train.shape[0])
```
ogrisel: Actually, it might even be smarter to combine the computation of the initial `gradients`, `hessians` and baseline `y_pred` in a single method as it's likely that they can share some intermediate computation.

```python
        y_pred, gradients, hessians = self.loss_.baseline_predictions(y_train)
```

Again, this should better be addressed in a separate PR.

pygbm/gradient_boosting.py
```diff
@@ -196,8 +214,56 @@ def _should_stop(self, scores):
...
+            loss=loss, learning_rate=learning_rate, max_iter=max_iter,
+            max_leaf_nodes=max_leaf_nodes, max_depth=max_depth,
+            min_samples_leaf=min_samples_leaf,
+            l2_regularization=l2_regularization, max_bins=max_bins,
+            max_no_improvement=max_no_improvement,
+            validation_split=validation_split, scoring=scoring, tol=tol,
+            verbose=verbose, random_state=random_state)
+
+    def predict(self, X):
+        return self._decision_function(X)
```
ogrisel: It's a bit weird to call a method named `_decision_function` in a regressor. Maybe it should still be called `predict` in the base class so that it can be inherited directly in the regressor and a be overriden by a loss specific method in the classifier subclass. Maybe some regression loss function would also need to have the loss be able to post process the raw predictions of the trees.

Alternatively call it `_raw_predict` in the base class.

pygbm/gradient_boosting.py
ogrisel: I think we should do that now as part of this PR.

pygbm/loss.py
```diff
@@ -0,0 +1,97 @@
...
+    # returns the starts and ends of each region. Used to simulate a 'static'
+    # scheduling.
+    n_threads = numba.config.NUMBA_DEFAULT_NUM_THREADS
+    sizes = np.full(n_threads, total_size // n_threads, dtype=np.int32)
+    sizes[:total_size % n_threads] += 1
+    starts = np.zeros(n_threads, dtype=np.int32)
+    starts[1:] = np.cumsum(sizes[:-1])
+    ends = starts + sizes
+
+    return starts, ends, n_threads
```
ogrisel: It would be great to contact the numba core devs to get their opinion on this kind of threading-scheduling boiler plate code.

Also, shouldn't we use `n_chunks = 3 * n_threads` to mitigate potential stragglers?

ogrisel: In particular it would be great if numba could provided some kind of ["guided" scheduling strategy](http://jakascorner.com/blog/2016/06/omp-for-scheduling.html#guided) like openmp does, in particular with the ability to pass a minimum chunk size.

NicolasHug: I commented https://github.com/numba/numba/issues/2649#issuecomment-442200409 about this a few days ago

pygbm/gradient_boosting.py
NicolasHug: That's what scikitlearn is doing but I agree, that was actually a question I had for you in #49.

I proposed `_get_scores()` but I like `_raw_predict()` as well

NicolasHug:
I addressed the comments.

Haven't done any progress regarding the newton test :/ I tried with float64 but  I'm still getting the convergence error

ogrisel:
I think we should make it explicit in the loss API that `y_pred` from the trees is actually a raw score and needs to go through the link function before being able to interpret it and compare it to `y_train`. For least squares, `y_pred` and `y_train` are homogeneous and can be directly compared but this is not the case for other loss function in general. This is probably the cause of the newton test failure. I will try to dig a bit further.

ogrisel:
Maybe we could rename `y_pred` to `predicted_scores` internally and always have:

```python
y_pred = self.loss_.inverse_link_function(predicted_scores)
```

NicolasHug:
Yes I agree 100% I was planning to submit a PR about this. But I don't think this causes the issue because y_pred in __call__ is homogeneous to that in update_gradients_and_hessians.

Else the numerical comparison test would fail.

ogrisel:
Ok my loss functions seems to break the numerical gradient check while yours passes...

ogrisel:
I messed up with git and lost my changes. I have to go offline for approximately two hours. Feel free to takeover if you wish. Otherwise I will give it another try.

NicolasHug:
Won't be there either ^^

NicolasHug:
I'm not sure if I understand your loss formulation. Is it for a -1 / 1 encoding?

The one I'm using is from [The Elements of
Statistical Learning](https://web.stanford.edu/~hastie/ElemStatLearn/]) section 4.4.1

`logistic loss = - log likelihood = sum_x log(proba that x belongs to its true class)`

The proba that x belongs to class 1 is defined as `p_1(x) = 1 - p_0(x) = sigmoid(y_pred)`
where y_pred is the raw values from the tree.

A bit a calculus leads to the unifying expression used in the code:

`log(proba that x belongs to its true class)` = `log(1 + e(y_pred)) - y * y_pred`

ogrisel:
I have the fix working. I will push it.

ogrisel:
I proposed my loss from memory without checking but obviously it's wrong because the numerical gradient test did fail. Your is correct.

ogrisel:
The warnings in the newton test should disappear with the upcoming scipy 1.2.0 release: https://github.com/scipy/scipy/pull/8907.

ogrisel:
@NicolasHug Indeed your analysis is right, the expression I used is only when y_true takes values in `{-1, 1}` instead of `{0, 1}`.

pygbm/loss.py
```diff
@@ -0,0 +1,108 @@
...
+    n_samples = gradients.shape[0]
+    starts, ends, n_threads = _get_threads_chunks(total_size=n_samples)
+    for thread_idx in prange(n_threads):
+        for i in range(starts[thread_idx], ends[thread_idx]):
+            gradients[i] = _expit(y_pred[i]) - y_true[i]
+            gradient_abs = np.abs(gradients[i])
+            hessians[i] = gradient_abs * (1. - gradient_abs)
+
+
+_LOSSES = {'ls': LeastSquares,
```
ogrisel: I would also be in favor of expanding `"ls"` to `"least_squares"`.

NicolasHug:
Thanks a lot for fixing the test!

+1 for `binary_crossentropy` and `categorical_crossentropy`

How about also accepting `crossentropy` which would result in either one depending on the nature of the problem?

ogrisel:
Technically, least squares is also cross-entropy aka NLL with a Gaussian model resulting in the identity link function in Generalized Linear Model parlance.

It's true that it would be convenient to have a default value that switches between the binomial and categorical model depending on the number of unique classes observed in the training set. However I would rather put `loss="auto"` in the constructor and do the switch in the `fit` method of the classifier.

NicolasHug:
Thanks a lot for the help @ogrisel I'll merge once the checks go green and go on with #49  after rebasing

PR #14, opened 2018-10-04
[MRG] Histogram computation optimization

NicolasHug:
This is an attempt to use the relation

`hist(parent) = hist(parent.left) + hist(parent.right)`

to only compute explicitly either `hist(parent.left)` or `hist(parent.right)`, and get the sibling's histogram by a simple subtraction.

This is supposed to be faster because explicitly computing a histogram is `O(n_samples_at_node)` while the subtraction is only `O(n_bins)` and `n_samples_at_node >> n_bins`.

TODO:

- [X] Ideally only the histogram of the sibling with the least number of samples should be explicitly computed. The current implementation does not support this and will compute the histogram of whatever node comes first during the tree growing process.
- [ ] The code is a bit ugly, could be refactored into something better
- [X] Make it actually work...
- [x] Write some tests

The good news is, the predictions are the same as before. ~~The bad news is, no clear performance gain is achieved... I'll try to understand what's going on.~~


ogrisel:
> Ideally only the histogram of the sibling with the least number of samples should be explicitly computed. The current implementation does not support this and will compute the histogram of whatever node comes first during the tree growing process.

This can be fixed easily in the grower itself by choosing to call compute split-ability of the smallest node first.

ogrisel:
> The bad news is, no clear performance gain is achieved... I'll try to understand what's going on.

Too bad. You might need to add some `start_time = time()` / `duration += time() - start_time` manual profiling probes around the interesting parts to understand what's going on.

NicolasHug:
>This can be fixed easily in the grower itself by choosing to call compute split-ability of the smallest node first.

So I tried doing just that, here's a tree example.
digraph

`fast` means that the node histogram was computed with the histogram trick, `time` corresponds to the time needed to compute the histogram + the best split, and ratio corresponds to `(node.sibling.time) / (node.time)` when the sibling's histogram was computed with the slow method, and `1` otherwise. So it gives an idea of how much faster the fast method is.

Now something pretty weird is going on: I computed the average ratio of the `fast` nodes:

```
mean_ratio = np.mean([node['ratio'] for pred in pygbm_model.predictors_
                      for node in pred.nodes if node['ratio'] != 1])
```

and this value is < 1 when we use 'fast' on the siblings with the highest number of samples (which is supposed to be the correct way). Also, it's > 1 if we use the 'fast' method on the siblings with the least number of samples. This might be due to the fact that `time` does not correspond to the time needed to only compute the histogram, but also to compute the best split.




Other thoughts:

- ordered_gradients and ordered_hessians are not needed when for the fast. We're computing them in all cases so that's a waste of time. That might also explain those discrepancies.
- When computing the hist of a sibling because it has a fewer number of samples,  if that sibling is a leaf (or will become one) we don't need to compute the splittability, only the hist. We're currently wasting time in computing the splittability everytime.

ogrisel:
You should replot the above analysis on a larger dataset such as the Higgs boson dataset with 1e7 samples. Otherwise, the splits are too quick and the timing probes might induce some non-trivial overhead. To make it easier to debug, use a smaller tree with a 5 leaf nodes limit for instance

In the above graph, only the first level splits with large sample counts benefit from the optimization.

ogrisel:
Based on your tree, it seems that when count < 4e5 samples, it's always faster to compute the histogram (<1ms). Doing the histogram substraction is slower: >1ms.

So I think you should enable this histogram substraction if the node count is larger than 5e5 as a rule of thumb.

ogrisel:
Actually my analysis is wrong because it depends on the imbalance between the sibling nodes. Doing the histogram optimization probably never hurts but it's not expected to yield huge improvements when the count are below 5e5.

In your plot, think it would be more interesting to report "count / time" values and the ratio of (count_fast / time_fast)  / (count_slow / time_slow).

ogrisel:
I think it would be interesting to build LightGBM with its TIMETAG feature enabled and collect similar statistics in pygbm to be able to compare.

https://github.com/Microsoft/LightGBM/blob/master/src/treelearner/serial_tree_learner.cpp#L34

It would also be interesting to count the number of times the histogram computation functions are called for each implementation to check that we are consistent.

ogrisel:
To enable TIMETAG, just edit the CMakeLists.txt at the root of the LightGBM source folder to add:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 057302b..b735793 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,9 @@ elseif(MSVC)
   cmake_minimum_required(VERSION 3.8)
 endif()
 
+
+add_definitions(-DTIMETAG)
+
 if(USE_SWIG)
   find_package(SWIG REQUIRED)
   find_package(Java REQUIRED)
```

the rebuild the python package: `cd python-package; python setup.py install`.

Passing verbose >= 2 should output the following (e.g. in the Higgs Boson benchmark):

```
[LightGBM] [Info] GBDT::boosting costs 0.188611
[LightGBM] [Info] GBDT::train_score costs 0.807898
[LightGBM] [Info] GBDT::out_of_bag_score costs 0.000005
[LightGBM] [Info] GBDT::valid_score costs 0.000001
[LightGBM] [Info] GBDT::metric costs 0.000000
[LightGBM] [Info] GBDT::bagging costs 0.000006
[LightGBM] [Info] GBDT::tree costs 7.310715
[LightGBM] [Info] SerialTreeLearner::init_train costs 0.122483
[LightGBM] [Info] SerialTreeLearner::init_split costs 0.000371
[LightGBM] [Info] SerialTreeLearner::hist_build costs 5.476519
[LightGBM] [Info] SerialTreeLearner::find_split costs 0.325841
[LightGBM] [Info] SerialTreeLearner::split costs 1.381069
[LightGBM] [Info] SerialTreeLearner::ordered_bin costs 0.000000
```

pygbm/histogram.py
```diff
@@ -20,6 +20,40 @@ def _build_ghc_histogram_naive(n_bins, sample_indices, binned_feature,
     return histogram
 
 
+@njit
+def _build_ghc_histogram_unrolled_fast(n_bins, parent_histogram,
+                                       sibling_histogram):
+    histogram = np.zeros(n_bins, dtype=HISTOGRAM_DTYPE)
+    unrolled_upper = (n_bins // 4) * 4
+
+    for i in range(0, unrolled_upper, 4):
```
ogrisel: Are you sure that unrolling bring any speed up for just a short iteration? n_bins is expected to be 255 max.

pygbm/histogram.py
```diff
@@ -20,6 +20,40 @@ def _build_ghc_histogram_naive(n_bins, sample_indices, binned_feature,
     return histogram
 
 
+@njit
+def _build_ghc_histogram_unrolled_fast(n_bins, parent_histogram,
+                                       sibling_histogram):
```
ogrisel: I would rename this a `_substract_histograms` and `parent_histogram` to `hist_a` and `sibling_histogram` to `hist_b`.

ogrisel: Also, I think that you should create a new method name `find_node_split_by_substraction` that would call a new function named `_parallel_substract_histograms`.

And keep `find_node_split` for the case where we actually compute histograms with `_parallel_find_splits`.

NicolasHug: Yes this definitely needs some refactoring. Thanks for all the comments, I'll go back to it soon.

Also, any idea on how to report execution time from inside numba? importing the `time` module does not work when in no-python mode. We're currently reporting time(hist + splittability) and what we're interested in is only time(hist).

pygbm/histogram.py
NicolasHug: No idea... I just thought it coudn't hurt but indeed I don't think it's really useful.

NicolasHug:
Soooo we're getting closer :)

I simply duplicated the whole pipeline, the only difference between calling `find_node_split_subtraction` and `find_node_split` is that `find_node_split_subtraction` will end up calling  `_subtract_ghc_histograms_unrolled` instead of one of the other 'low' histogram building functions.

And just like that, I get to run boson in 23 seconds (17 for lightgbm). That's pretty weird to me because the whole `find_node_split` didn't change at all, but I won't complain. Numba mysteries I suppose.  

Now, I tried to get rid of all the `ordered_gradient` stuff. As far as I understand, we should be able to compute the hessians and gradients of the current node by just [summing up the histogram](https://github.com/ogrisel/pygbm/pull/14/files#diff-43a3fcc0b6fe7d9cdd257e9b6d2b4fe3R342), right? Unfortunately this will break my basic sanity check (AUC is not the same as before). Any idea what might be going on? Or am I simply mistaken about the computation?

It's super worth it though, because once we get rid of the `ordered_stuff` computation, boson runs in 12 seconds ^^

Last comments:
- Code is super ugly but I just wanted your input
- sorry about the first WIP, you can ignore it
- no idea what's going on with travis, tests are passing locally

ogrisel:
> And just like that, I get to run boson in 23 seconds (17 for lightgbm). That's pretty weird to me because the whole find_node_split didn't change at all, but I won't complain. Numba mysteries I suppose.

Nice! It's possible that the numba type inference had trouble with the previous code layout and was missing additional typing hints.

It's always a good idea to check the type inferred for all the local variables of the performance critical sections of the code so as to ensure that there is not unnecessary type casts that might prevent the compiler to optimize the generated code:

https://github.com/ogrisel/pygbm/#debugging-numba-type-inference

> Now, I tried to get rid of all the ordered_gradient stuff. As far as I understand, we should be able to compute the hessians and gradients of the current node by just summing up the histogram, right? 

Yes I believe so. Maybe you could alternative try to store the gradient and hessian sums on the grower nodes themselves along with the histograms as ancillary arrays.

> It's super worth it though, because once we get rid of the ordered_stuff computation, boson runs in 12 seconds ^^

Sounds promising. Once we get good performance on a few cores, it would be interesting to check the scalability on hardware with many more cores (e.g. 32 on a cloud machine for instance) and contrast that to the performance profile of lightgbm.

Also note, the following test times out when disabling the numba jit: 

https://travis-ci.org/ogrisel/pygbm/jobs/439885373

Maybe it could be made faster so that it does not fail when numba is disabled. Running without numba is useful to collect coverage data (and also possibly to check that numba itself not introducing unexpected bugs :)

NicolasHug:
All right I just found what was wrong, it was the hessian computation:

`hessian_parent = np.sum(histogram[:]['sum_hessians'])`

this was run regardless of the `constant_hessian` value. As here the hessian is constant, the `sum_hessian` field of the histogram of the sibling (with the slow method) would never be computed and always be 0. So the current histogram would be wrong for the `sum_hessian` field (always zero as well), and we would always hit 

```
        if hessian_left < min_hessian_to_split:
            continue
```

>There will likely be some rounding errors

Indeed, there are rounding errors. The relative error between the `gradient_parent` variable computed with the 2 methods is always less than 1E-3 though, and usually much less. As far as I can tell the splits are still the same (trees have the same structure), but the value of the leaves slightly change:

```
>       assert roc_auc == 0.9809811751028725
E       assert 0.9809811974530592 == 0.9809811751028725

```

I'll clean this all up!! :)

NicolasHug:
Here's a first clean up, I'll push some more soon!

NicolasHug:
>It's super worth it though, because once we get rid of the ordered_stuff computation, boson runs in 12 seconds ^^

Unfortunately this was an side effect of the hessian bug: every node computed with the fast method would end up being a leaf, so trees where much smaller, hence the very fast training time...

Now I get to run boson in about the same time as LightGBM, but there's a strange behavior. When I first run the benchmark, pygbm is pretty slow, but the more I run the it the faster it gets, and it gets stable around 20s. LightGBM runtime is almost the same every time though. I did this 2 times, with about 1h between the 2 experiments, and I get the same thing in both. Any idea? Is there some strange caching thing going on?

<details>
```
~/dev/pygbm(branch:histogram_subtraction*) » python benchmarks/bench_higgs_boson.py
Training set with 10950000 records with 28 features.
Fitting a LightGBM model...
[LightGBM] [Info] Total Bins 6143
[LightGBM] [Info] Number of data: 10950000, number of used features: 28
[LightGBM] [Info] Start training from score 0.529920
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 7
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 10
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 9
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 11
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 15
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 12
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 13
done in 17.587s, ROC AUC: 0.7872
JIT compiling code for the pygbm model...
done in 4.635s
Fitting a pygbm model...
Binning 1.226 GB of data: 6.959 s (176.244 MB/s)
Fitting gradient boosted rounds:
[0/10]
[1/10] 31 leaf nodes, max depth 7 in 1.388s
[2/10] 31 leaf nodes, max depth 10 in 1.510s
[3/10] 31 leaf nodes, max depth 9 in 1.737s
[4/10] 31 leaf nodes, max depth 10 in 2.292s
[5/10] 31 leaf nodes, max depth 10 in 3.270s
[6/10] 31 leaf nodes, max depth 9 in 3.343s
[7/10] 31 leaf nodes, max depth 11 in 3.061s
[8/10] 31 leaf nodes, max depth 8 in 2.826s
[9/10] 31 leaf nodes, max depth 11 in 2.660s
[10/10] 31 leaf nodes, max depth 11 in 2.576s
Fit 10 trees in 32.908 s, (310 total leaf nodes)
done in 32.951s, ROC AUC: 0.7892
(pygbm) ------------------------------------------------------------
~/dev/pygbm(branch:histogram_subtraction*) » python benchmarks/bench_higgs_boson.py
Training set with 10950000 records with 28 features.
Fitting a LightGBM model...
[LightGBM] [Info] Total Bins 6143
[LightGBM] [Info] Number of data: 10950000, number of used features: 28
[LightGBM] [Info] Start training from score 0.529920
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 7
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 10
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 9
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 11
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 15
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 12
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 13
done in 17.592s, ROC AUC: 0.7872
JIT compiling code for the pygbm model...
done in 4.646s
Fitting a pygbm model...
Binning 1.226 GB of data: 6.896 s (177.832 MB/s)
Fitting gradient boosted rounds:
[0/10]
[1/10] 31 leaf nodes, max depth 7 in 1.377s
[2/10] 31 leaf nodes, max depth 10 in 1.348s
[3/10] 31 leaf nodes, max depth 9 in 1.379s
[4/10] 31 leaf nodes, max depth 10 in 1.950s
[5/10] 31 leaf nodes, max depth 10 in 2.659s
[6/10] 31 leaf nodes, max depth 9 in 2.528s
[7/10] 31 leaf nodes, max depth 11 in 2.350s
[8/10] 31 leaf nodes, max depth 8 in 2.204s
[9/10] 31 leaf nodes, max depth 11 in 2.107s
[10/10] 31 leaf nodes, max depth 11 in 2.015s
Fit 10 trees in 27.228 s, (310 total leaf nodes)
done in 27.267s, ROC AUC: 0.7892
(pygbm) ------------------------------------------------------------
~/dev/pygbm(branch:histogram_subtraction*) » python benchmarks/bench_higgs_boson.py
Training set with 10950000 records with 28 features.
Fitting a LightGBM model...
[LightGBM] [Info] Total Bins 6143
[LightGBM] [Info] Number of data: 10950000, number of used features: 28
[LightGBM] [Info] Start training from score 0.529920
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 7
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 10
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 9
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 11
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 15
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 12
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 13
done in 17.588s, ROC AUC: 0.7872
JIT compiling code for the pygbm model...
done in 4.629s
Fitting a pygbm model...
Binning 1.226 GB of data: 6.947 s (176.540 MB/s)
Fitting gradient boosted rounds:
[0/10]
[1/10] 31 leaf nodes, max depth 7 in 1.392s
[2/10] 31 leaf nodes, max depth 10 in 1.373s
[3/10] 31 leaf nodes, max depth 9 in 1.350s
[4/10] 31 leaf nodes, max depth 10 in 1.364s
[5/10] 31 leaf nodes, max depth 10 in 1.365s
[6/10] 31 leaf nodes, max depth 9 in 1.371s
[7/10] 31 leaf nodes, max depth 11 in 1.359s
[8/10] 31 leaf nodes, max depth 8 in 1.336s
[9/10] 31 leaf nodes, max depth 11 in 1.336s
[10/10] 31 leaf nodes, max depth 11 in 1.317s
Fit 10 trees in 20.283 s, (310 total leaf nodes)
done in 20.300s, ROC AUC: 0.7892
(pygbm) ------------------------------------------------------------
~/dev/pygbm(branch:histogram_subtraction*) » python benchmarks/bench_higgs_boson.py
Training set with 10950000 records with 28 features.
Fitting a LightGBM model...
[LightGBM] [Info] Total Bins 6143
[LightGBM] [Info] Number of data: 10950000, number of used features: 28
[LightGBM] [Info] Start training from score 0.529920
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 7
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 10
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 9
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 11
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 15
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 12
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 13
done in 17.268s, ROC AUC: 0.7872
JIT compiling code for the pygbm model...
done in 4.584s
Fitting a pygbm model...
Binning 1.226 GB of data: 6.954 s (176.362 MB/s)
Fitting gradient boosted rounds:
[0/10]
[1/10] 31 leaf nodes, max depth 7 in 1.393s
[2/10] 31 leaf nodes, max depth 10 in 1.376s
[3/10] 31 leaf nodes, max depth 9 in 1.359s
[4/10] 31 leaf nodes, max depth 10 in 1.373s
[5/10] 31 leaf nodes, max depth 10 in 1.365s
[6/10] 31 leaf nodes, max depth 9 in 1.370s
[7/10] 31 leaf nodes, max depth 11 in 1.353s
[8/10] 31 leaf nodes, max depth 8 in 1.331s
[9/10] 31 leaf nodes, max depth 11 in 1.333s
[10/10] 31 leaf nodes, max depth 11 in 1.313s
Fit 10 trees in 20.255 s, (310 total leaf nodes)
done in 20.275s, ROC AUC: 0.7892
(pygbm) ------------------------------------------------------------
~/dev/pygbm(branch:histogram_subtraction*) » python benchmarks/bench_higgs_boson.py
Training set with 10950000 records with 28 features.
Fitting a LightGBM model...
[LightGBM] [Info] Total Bins 6143
[LightGBM] [Info] Number of data: 10950000, number of used features: 28
[LightGBM] [Info] Start training from score 0.529920
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 7
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 10
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 9
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 11
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 15
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 12
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 13
done in 17.203s, ROC AUC: 0.7872
JIT compiling code for the pygbm model...
done in 4.571s
Fitting a pygbm model...
Binning 1.226 GB of data: 6.899 s (177.774 MB/s)
Fitting gradient boosted rounds:
[0/10]
[1/10] 31 leaf nodes, max depth 7 in 1.372s
[2/10] 31 leaf nodes, max depth 10 in 1.360s
[3/10] 31 leaf nodes, max depth 9 in 1.343s
[4/10] 31 leaf nodes, max depth 10 in 1.355s
[5/10] 31 leaf nodes, max depth 10 in 1.348s
[6/10] 31 leaf nodes, max depth 9 in 1.355s
[7/10] 31 leaf nodes, max depth 11 in 1.338s
[8/10] 31 leaf nodes, max depth 8 in 1.318s
[9/10] 31 leaf nodes, max depth 11 in 1.320s
[10/10] 31 leaf nodes, max depth 11 in 1.303s
Fit 10 trees in 20.095 s, (310 total leaf nodes)
done in 20.109s, ROC AUC: 0.7892
```
</details>


<details>

```
~/dev/pygbm(branch:histogram_subtraction*) » python benchmarks/bench_higgs_boson.py                                                                                                nico@cotier
Training set with 10950000 records with 28 features.
Fitting a LightGBM model...
[LightGBM] [Info] Total Bins 6143
[LightGBM] [Info] Number of data: 10950000, number of used features: 28
[LightGBM] [Info] Start training from score 0.529920
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 7
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 10
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 9
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 11
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 15
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 12
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 13
done in 17.350s, ROC AUC: 0.7872
JIT compiling code for the pygbm model...
done in 4.602s
Fitting a pygbm model...
Binning 1.226 GB of data: 6.966 s (176.066 MB/s)
Fitting gradient boosted rounds:
[0/10]
[1/10] 31 leaf nodes, max depth 7 in 1.374s
[2/10] 31 leaf nodes, max depth 10 in 1.389s
[3/10] 31 leaf nodes, max depth 9 in 1.382s
[4/10] 31 leaf nodes, max depth 10 in 4.190s
[5/10] 31 leaf nodes, max depth 10 in 4.339s
[6/10] 31 leaf nodes, max depth 9 in 4.082s
[7/10] 31 leaf nodes, max depth 11 in 3.686s
[8/10] 31 leaf nodes, max depth 8 in 3.376s
[9/10] 31 leaf nodes, max depth 11 in 3.159s
[10/10] 31 leaf nodes, max depth 11 in 3.078s
Fit 10 trees in 37.930 s, (310 total leaf nodes)
done in 37.980s, ROC AUC: 0.7892
(pygbm) ------------------------------------------------------------
~/dev/pygbm(branch:histogram_subtraction*) » python benchmarks/bench_higgs_boson.py
Training set with 10950000 records with 28 features.
Fitting a LightGBM model...
[LightGBM] [Info] Total Bins 6143
[LightGBM] [Info] Number of data: 10950000, number of used features: 28
[LightGBM] [Info] Start training from score 0.529920
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 7
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 10
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 9
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 11
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 15
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 12
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 13
done in 17.477s, ROC AUC: 0.7872
JIT compiling code for the pygbm model...
done in 4.568s
Fitting a pygbm model...
Binning 1.226 GB of data: 6.986 s (175.542 MB/s)
Fitting gradient boosted rounds:
[0/10]
[1/10] 31 leaf nodes, max depth 7 in 1.373s
[2/10] 31 leaf nodes, max depth 10 in 1.363s
[3/10] 31 leaf nodes, max depth 9 in 1.351s
[4/10] 31 leaf nodes, max depth 10 in 1.383s
[5/10] 31 leaf nodes, max depth 10 in 2.467s
[6/10] 31 leaf nodes, max depth 9 in 2.470s
[7/10] 31 leaf nodes, max depth 11 in 2.299s
[8/10] 31 leaf nodes, max depth 8 in 2.160s
[9/10] 31 leaf nodes, max depth 11 in 2.072s
[10/10] 31 leaf nodes, max depth 11 in 2.000s
Fit 10 trees in 27.157 s, (310 total leaf nodes)
done in 27.191s, ROC AUC: 0.7892
(pygbm) ------------------------------------------------------------
~/dev/pygbm(branch:histogram_subtraction*) » python benchmarks/bench_higgs_boson.py
Training set with 10950000 records with 28 features.
Fitting a LightGBM model...
[LightGBM] [Info] Total Bins 6143
[LightGBM] [Info] Number of data: 10950000, number of used features: 28
[LightGBM] [Info] Start training from score 0.529920
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 7
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 10
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 9
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 11
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 15
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 12
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 13
done in 17.416s, ROC AUC: 0.7872
JIT compiling code for the pygbm model...
done in 4.571s
Fitting a pygbm model...
Binning 1.226 GB of data: 6.899 s (177.756 MB/s)
Fitting gradient boosted rounds:
[0/10]
[1/10] 31 leaf nodes, max depth 7 in 1.369s
[2/10] 31 leaf nodes, max depth 10 in 1.360s
[3/10] 31 leaf nodes, max depth 9 in 1.338s
[4/10] 31 leaf nodes, max depth 10 in 1.355s
[5/10] 31 leaf nodes, max depth 10 in 1.352s
[6/10] 31 leaf nodes, max depth 9 in 1.355s
[7/10] 31 leaf nodes, max depth 11 in 1.337s
[8/10] 31 leaf nodes, max depth 8 in 1.315s
[9/10] 31 leaf nodes, max depth 11 in 1.320s
[10/10] 31 leaf nodes, max depth 11 in 1.301s
Fit 10 trees in 20.077 s, (310 total leaf nodes)
done in 20.097s, ROC AUC: 0.7892
(pygbm) ------------------------------------------------------------
~/dev/pygbm(branch:histogram_subtraction*) » python benchmarks/bench_higgs_boson.py
Training set with 10950000 records with 28 features.
Fitting a LightGBM model...
[LightGBM] [Info] Total Bins 6143
[LightGBM] [Info] Number of data: 10950000, number of used features: 28
[LightGBM] [Info] Start training from score 0.529920
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 7
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 10
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 9
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 11
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 15
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 12
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 13
done in 17.178s, ROC AUC: 0.7872
JIT compiling code for the pygbm model...
done in 4.542s
Fitting a pygbm model...
Binning 1.226 GB of data: 6.909 s (177.511 MB/s)
Fitting gradient boosted rounds:
[0/10]
[1/10] 31 leaf nodes, max depth 7 in 1.371s
[2/10] 31 leaf nodes, max depth 10 in 1.348s
[3/10] 31 leaf nodes, max depth 9 in 1.328s
[4/10] 31 leaf nodes, max depth 10 in 1.344s
[5/10] 31 leaf nodes, max depth 10 in 1.340s
[6/10] 31 leaf nodes, max depth 9 in 1.339s
[7/10] 31 leaf nodes, max depth 11 in 1.331s
[8/10] 31 leaf nodes, max depth 8 in 1.312s
[9/10] 31 leaf nodes, max depth 11 in 1.315s
[10/10] 31 leaf nodes, max depth 11 in 1.297s
Fit 10 trees in 20.040 s, (310 total leaf nodes)
done in 20.057s, ROC AUC: 0.7892
(pygbm) ------------------------------------------------------------
~/dev/pygbm(branch:histogram_subtraction*) » python benchmarks/bench_higgs_boson.py
Training set with 10950000 records with 28 features.
Fitting a LightGBM model...
[LightGBM] [Info] Total Bins 6143
[LightGBM] [Info] Number of data: 10950000, number of used features: 28
[LightGBM] [Info] Start training from score 0.529920
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 7
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 10
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 9
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 11
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 8
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 15
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 12
[LightGBM] [Debug] Trained a tree with leaves = 31 and max_depth = 13
done in 17.192s, ROC AUC: 0.7872
JIT compiling code for the pygbm model...
done in 4.581s
Fitting a pygbm model...
Binning 1.226 GB of data: 6.912 s (177.429 MB/s)
Fitting gradient boosted rounds:
[0/10]
[1/10] 31 leaf nodes, max depth 7 in 1.359s
[2/10] 31 leaf nodes, max depth 10 in 1.346s
[3/10] 31 leaf nodes, max depth 9 in 1.331s
[4/10] 31 leaf nodes, max depth 10 in 1.349s
[5/10] 31 leaf nodes, max depth 10 in 1.343s
[6/10] 31 leaf nodes, max depth 9 in 1.340s
[7/10] 31 leaf nodes, max depth 11 in 1.351s
[8/10] 31 leaf nodes, max depth 8 in 1.338s
[9/10] 31 leaf nodes, max depth 11 in 1.349s
[10/10] 31 leaf nodes, max depth 11 in 1.331s
Fit 10 trees in 20.392 s, (310 total leaf nodes)
done in 20.408s, ROC AUC: 0.7892


```
</details>

NicolasHug:
Just FWI I've opened https://github.com/numba/numba/issues/3405 regarding one of the refactoring issue.

tests/test_histogram.py
```diff
@@ -110,3 +117,85 @@ def test_compare_histograms_optimized_vs_general():
...
+                                           binned_feature,
+                                           ordered_gradients)
+    hist_ghc = _build_ghc_histogram_unrolled(n_bins, sample_indices,
+                                             binned_feature,
+                                             ordered_gradients,
+                                             ordered_hessians)
+
+    hist_naive = _build_ghc_histogram_naive(n_bins, sample_indices,
+                                            binned_feature, ordered_gradients,
+                                            ordered_hessians)
```
ogrisel: Because we use the unrolled we could rename all the `_build_*_histogram_unrolled` to just `_build_*_histogram` but keep the "naive" suffix in `_build_ghc_histogram_naive`.

Also the `gc` / `ghc` stuff is confusing.

Maybe:

- `_build_ghc_root_histogram_unrolled` =>`_build_histogram_root_node`;
- `_build_gc_root_histogram_unrolled` =>`_build_histogram_root_node_no_hessian`;
- `_build_ghc_histogram_unrolled` =>`_build_histogram`;
- `_build_gc_histogram_unrolled` =>`_build_histogram_no_hessian`.
- `_build_ghc_histogram_naive` => `_build_histogram_naive`

tests/test_histogram.py
```diff
@@ -110,3 +117,85 @@ def test_compare_histograms_optimized_vs_general():
...
+                                               ordered_gradients_right,
+                                               ordered_hessians_right)
+
+    hist_left_sub = _subtract_ghc_histograms_unrolled(n_bins, hist_parent,
+                                                      hist_right)
+    hist_right_sub = _subtract_ghc_histograms_unrolled(n_bins, hist_parent,
+                                                      hist_left)
+
+    for key in ('count', 'sum_hessians', 'sum_gradients'):
+        assert_allclose(hist_left[key], hist_left_sub[key], rtol=1e-6)
```
ogrisel: Can't you compare the full record array at once?

```python
assert_allclose(hist_left, hist_left_sub, rtol=1e-6)
```

or did you intentionally use fine grained comparison on each key individually to get more meaningful error reports in case of assertion failures?

tests/test_histogram.py
```diff
@@ -110,3 +117,85 @@ def test_compare_histograms_optimized_vs_general():
...
+    for hist in (hist_gc_root, hist_ghc_root, hist_gc, hist_gc, hist_ghc):
+        assert_array_equal(hist['count'], hist_naive['count'])
+        assert_allclose(hist['sum_gradients'], hist_naive['sum_gradients'])
+    for hist in (hist_ghc_root, hist_ghc):
+        assert_allclose(hist['sum_hessians'], hist_naive['sum_hessians'])
+    for hist in (hist_gc_root, hist_gc):
+        assert_array_equal(hist['sum_hessians'], np.zeros(n_bins))
+
+
+def test_hist_subtraction():
```
ogrisel: Please parametrize this test to test both the constant hessian and a non-constant hessian case:

```python
@pytest.mark.parametrize("constant_hessian", [True, False]):
def test_hist_subtraction(constant_hessian):
    ...
    if constant_hessian:
        ordered_hessians = np.ones(n_samples, dtype=np.float32)
    else:
        ordered_hessians = rng.lognormal(size=n_samples).astype(np.float32)
    ...
```

and when `constant_hessian` is `True` use the `no_hessian` variants of the `_build_histogram` methods both for the parent and the sibling data scanning computations.

tests/test_histogram.py
NicolasHug: I tried but I'm getting this error:
<details>
```
    
>       assert_allclose(hist_left, hist_left_sub, rtol=1e-6)

tests/test_histogram.py:140: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../.virtualenvs/pygbm/lib/python3.7/site-packages/numpy/testing/_private/utils.py:1447: in compare
    equal_nan=equal_nan)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

a = array([(-0.61542976, 1., 1), (-2.1266346 , 2., 2), (-0.34148723, 2., 2),
       ( 2.2112024 , 3., 3), ( 0.        , 0., 0)],
      dtype=[('sum_gradients', '<f4'), ('sum_hessians', '<f4'), ('count', '<u4')])
b = array([(-0.61542976, 1., 1), (-2.1266346 , 2., 2), (-0.34148723, 2., 2),
       ( 2.2112024 , 3., 3), ( 0.        , 0., 0)],
      dtype=[('sum_gradients', '<f4'), ('sum_hessians', '<f4'), ('count', '<u4')])
rtol = 1e-06, atol = 0, equal_nan = True

    def isclose(a, b, rtol=1.e-5, atol=1.e-8, equal_nan=False):
        """
        Returns a boolean array where two arrays are element-wise equal within a
        tolerance.
    
        The tolerance values are positive, typically very small numbers.  The
        relative difference (`rtol` * abs(`b`)) and the absolute difference
        `atol` are added together to compare against the absolute difference
        between `a` and `b`.
    
        .. warning:: The default `atol` is not appropriate for comparing numbers
                     that are much smaller than one (see Notes).
    
        Parameters
        ----------
        a, b : array_like
            Input arrays to compare.
        rtol : float
            The relative tolerance parameter (see Notes).
        atol : float
            The absolute tolerance parameter (see Notes).
        equal_nan : bool
            Whether to compare NaN's as equal.  If True, NaN's in `a` will be
            considered equal to NaN's in `b` in the output array.
    
        Returns
        -------
        y : array_like
            Returns a boolean array of where `a` and `b` are equal within the
            given tolerance. If both `a` and `b` are scalars, returns a single
            boolean value.
    
        See Also
        --------
        allclose
    
        Notes
        -----
        .. versionadded:: 1.7.0
    
        For finite values, isclose uses the following equation to test whether
        two floating point values are equivalent.
    
         absolute(`a` - `b`) <= (`atol` + `rtol` * absolute(`b`))
    
        Unlike the built-in `math.isclose`, the above equation is not symmetric
        in `a` and `b` -- it assumes `b` is the reference value -- so that
        `isclose(a, b)` might be different from `isclose(b, a)`. Furthermore,
        the default value of atol is not zero, and is used to determine what
        small values should be considered close to zero. The default value is
        appropriate for expected values of order unity: if the expected values
        are significantly smaller than one, it can result in false positives.
        `atol` should be carefully selected for the use case at hand. A zero value
        for `atol` will result in `False` if either `a` or `b` is zero.
    
        Examples
        --------
        >>> np.isclose([1e10,1e-7], [1.00001e10,1e-8])
        array([True, False])
        >>> np.isclose([1e10,1e-8], [1.00001e10,1e-9])
        array([True, True])
        >>> np.isclose([1e10,1e-8], [1.0001e10,1e-9])
        array([False, True])
        >>> np.isclose([1.0, np.nan], [1.0, np.nan])
        array([True, False])
        >>> np.isclose([1.0, np.nan], [1.0, np.nan], equal_nan=True)
        array([True, True])
        >>> np.isclose([1e-8, 1e-7], [0.0, 0.0])
        array([ True, False], dtype=bool)
        >>> np.isclose([1e-100, 1e-7], [0.0, 0.0], atol=0.0)
        array([False, False], dtype=bool)
        >>> np.isclose([1e-10, 1e-10], [1e-20, 0.0])
        array([ True,  True], dtype=bool)
        >>> np.isclose([1e-10, 1e-10], [1e-20, 0.999999e-10], atol=0.0)
        array([False,  True], dtype=bool)
        """
        def within_tol(x, y, atol, rtol):
            with errstate(invalid='ignore'):
                return less_equal(abs(x-y), atol + rtol * abs(y))
    
        x = asanyarray(a)
        y = asanyarray(b)
    
        # Make sure y is an inexact type to avoid bad behavior on abs(MIN_INT).
        # This will cause casting of x later. Also, make sure to allow subclasses
        # (e.g., for numpy.ma).
>       dt = multiarray.result_type(y, 1.)
E       TypeError: invalid type promotion

../../.virtualenvs/pygbm/lib/python3.7/site-packages/numpy/core/numeric.py:2360: TypeError
```
</details>

I didn't really looked into it TBH...

ogrisel: Alright then.

tests/test_histogram.py
```diff
@@ -83,30 +83,80 @@ def test_compare_histograms_optimized_vs_general():
...
+
+    if constant_hessian:
+        hist_parent = _build_histogram_no_hessian(n_bins, sample_indices,
+                                                  binned_feature,
+                                                  ordered_gradients)
+    else:
+        hist_parent = _build_histogram(n_bins, sample_indices, binned_feature,
+                                       ordered_gradients, ordered_hessians)
+
+
```
ogrisel: Style: please do not insert double blank lines in the body of a function.

tests/test_splitting.py
```diff
@@ -23,8 +25,72 @@ def test_hitstogram_split(n_bins):
...
+
+    split_info_left_sub, hists_left_sub = splitter.find_node_split_subtraction(sample_indices_left, hists_parent, hists_right)
+    split_info_right_sub, hists_right_sub = splitter.find_node_split_subtraction(sample_indices_right, hists_parent, hists_left)
+    
+    for hists, hists_sub in ((hists_left, hists_left_sub),
+                             (hists_right, hists_right_sub)):
+        for hist, hist_sub in zip(hists, hists_sub):
+            for key in ('count', 'sum_hessians', 'sum_gradients'):
+                assert_allclose(hist[key], hist_sub[key], rtol=1e-3)
+
```
ogrisel: Same remark here.

NicolasHug:
I'll fix the currently failing test.

In the mean time, I don't think we can refactor much more because of numba/numba#3405. I think it might be worth it to try to flatten the current `HistogramSplitter` jitclass into simple functions, and use instead a struct / jitclass without methods, e.g. `SplitterContext`, that would serve as a blob of data passed to the different specialized functions. Would you consider it, or do we really want to keep the `HistogramSplitter` class?

ogrisel:
I am fine but the context is has two sets of attributes:

- the attributes that depend on a feature, currently stored as attributes of HistogramSplitter.
- the attributes that depend on a particular grower node such as `sample_indices`, `parent_histograms`, and `sibling_histograms`. 

So if you do new structs / jitclasses + pure functions to get rid of the `HistogramSplitter` jitclass I would introduce 2 splitting contexts:

- FeatureSplittingContext with the attributes of the current `HistogramSplitter`
- NodeSplittingContext with the node specific attributes:  `sample_indices`, `parent_histograms`, and `sibling_histograms`.

Then we can use pure functions and use the `parallel=True` numba annotation without restriction.

However I am worried about the performance overhead of new jitclass instanciations (one per grower node). Maybe `NodeSplittingContext` is useless and we can pass those 3 arguments manually to the `find_node_split` function.

Also, `find_node_split` should still be decorated with `@njit` for performance reasons. If that's the case that will not make it possible to workaround the limitation of numba/numba#3405 as commented in https://github.com/numba/numba/issues/3405#issuecomment-429885169.

NicolasHug:
>Also, find_node_split should still be decorated with @njit for performance reasons. If that's the case that will not make it possible to workaround the limitation of numba/numba#3405 as commented in numba/numba#3405 (comment).

Ha, true, I thought it was because of the jitclass but you're right it's more specifically about the nopython mode. We should probably postpone the refactoring until it's fixed then

pygbm/splitting.py
```diff
@@ -110,95 +96,212 @@ def find_node_split(self, sample_indices):
...
+    # Allocate the structure for the best split information. It can be
+    # returned as such (with a negative gain) if the min_hessian_to_split
+    # condition is not satisfied. Such invalid splits are later discarded by
+    # the TreeGrower.
+    best_split = SplitInfo(-1., 0, 0, 0., 0., 0., 0.)
+
     gradient_left, hessian_left = 0., 0.
-    for bin_idx in range(histogram.shape[0]):
+    for bin_idx in range(n_bins):
         gradient_left += histogram[bin_idx]['sum_gradients']
```
NicolasHug: I have a weird feeling about this `+=` sign. The histogram is already a cumulative sum as far as I understand, so why do we want a `+=` sign here? I feel like it should be `=`.

Maybe this is related to my previous comment: https://github.com/ogrisel/pygbm/pull/14#issuecomment-430034908

ogrisel: `histogram[bin_idx]['sum_gradients']` is the sum of the gradients w.r.t. the samples that fell precisely in `bin_idx` (as done in `_build_histogram`. `gradient_left` is the sum of gradients w.r.t. the samples of any bin <= `bin_idx`.

So I believe the current code with `+=` is correct.

tests/test_histogram.py
```diff
@@ -83,30 +83,79 @@ def test_compare_histograms_optimized_vs_general():
...
+    mask = rng.randint(0, 2, n_samples).astype(np.bool)
+
+    sample_indices_left = sample_indices[mask]
+    ordered_gradients_left = ordered_gradients[mask]
+    ordered_hessians_left = ordered_hessians[mask]
+    if constant_hessian:
+        hist_left = _build_histogram_no_hessian(n_bins, sample_indices_left,
+                                                binned_feature,
+                                                ordered_gradients_left)
```
ogrisel: Maybe you can also construct `hist_left_2 = _build_histogram(..., ordered_hessians_left)` and check that `hist_left` is all close to `hist_left_2`.

tests/test_histogram.py
```diff
@@ -83,30 +83,79 @@ def test_compare_histograms_optimized_vs_general():
...
+                                     binned_feature, ordered_gradients_left,
+                                     ordered_hessians_left)
+
+    sample_indices_right = sample_indices[~mask]
+    ordered_gradients_right = ordered_gradients[~mask]
+    ordered_hessians_right = ordered_hessians[~mask]
+    if constant_hessian:
+        hist_right = _build_histogram_no_hessian(n_bins, sample_indices_right,
+                                                 binned_feature,
+                                                 ordered_gradients_right)
```
ogrisel: same here: `hist_right_2 = _build_histogram(..., ordered_hessians_right)` and check that `hist_right` is all close to `hist_right_2`.

pygbm/splitting.py
NicolasHug: Ok thanks, I thought maybe `histogram[bin_idx]['sum_gradients']` was already the sum of gradients w.r.t. the samples of any bin <= bin_idx.

So yes the current code must be correct, but then the gradient at the current node should indeed be `sum(histogram[:]['sum_gradients'])`. Something fishy is going on (at least in my head ^^), I'll try to add some tests and identify a potential mistake I made.

tests/test_histogram.py
NicolasHug: I have parametrized `test_unrolled_equivalent_to_naive`, this test should now be covered here (if I correctly understand what you want)

ogrisel: Indeed. Thanks.

NicolasHug:
>Something fishy is going on (at least in my head ^^), I'll try to add some tests and identify a potential mistake I made.

I've added some tests and everything looks fine. I didn't have to fix anything so the fishyness was probably just me not wrapping my head around it.

I've moved back the computation of gradients and hessians into `find_node_split_subtraction()`. Now both pipeline are really similar.

NicolasHug:
Is there anything I should do w.r.t. the failing tests with `NUMBA_DISABLE_JIT="1"` ?

ogrisel:
> Is there anything I should do w.r.t. the failing tests with NUMBA_DISABLE_JIT="1" ?

Yes: remove it :)

More seriously, it's just way to slow: use a much smaller dataset. You cannot expect the pure python version to do large compute heavy tasks in a time limited test.

ogrisel:
The new version of the code seems a bit slower than previously: I get 21+s for pygbm on the higgs boson benchmark (vs 18s for lightgbm).

ogrisel:
> Is it consistently slower than before? Because I get a few variations between each tries. 

It's not that a big deal. The difference is close to the noise level of repeated runs. We can optimize that later.

I opened an issue in numba/numba#3417 to track this. In the mean time ok to use more functions then.

Let's remove the test_sanity test and merge this PR.

pygbm/grower.py
```diff
@@ -82,18 +95,65 @@ def grow(self):
...
+                split_info, histograms = (
+                    self.splitter.find_node_split_subtraction(
+                        node.sample_indices, node.parent.histograms,
+                        node.sibling.histograms)
+                )
+            else:
+                split_info, histograms = self.splitter.find_node_split(
+                    node.sample_indices)
+            toc = time()
+            node.time = toc - tic
```
ogrisel: I think we should also accumulate the the total time spent finding the best splits in a grower level attribute and the same for the time spent in applying the splits.

In the gradient boosting class we should also record the time spent in binning and finally in computing the predictions to get the new gradients and hessians for the next boosting iteration.

At the end of the fit of the gradient boosting, if verbose > 0 we should print a profiling report with the accumulated time for each of those type of operations in a similar way as LightGBM does.

pygbm/predictor.py
```diff
@@ -13,6 +13,11 @@
     ('right', np.uint32),
     ('gain', np.float32),
     ('depth', np.uint32),
+    ('use_sub', np.bool),
+    ('time', np.float32),
+    ('ratio', np.float32),
+    ('sum_g', np.float32),
+    ('sum_h', np.float32),
```
ogrisel: Do we really need  to store the sum of gradients and hessians in the predictor? Maybe we should find a way to just plot the grower tree instead of the predictor tree if the goal is to present a tree-structured profile report of the cost of growing a tree.

The predictor tree is part of the public API of the model and I would rather note pollute it with attributes that are only useful for pygbm developers who try to optimize the implementation.

NicolasHug: No we don't need those. Same goes for `time` and `ratio`, etc. Those were only useful for debugging.

I agree that we should instead plot the grower tree. That's quite easy to do I think, we would just need to keep a `growers_` list attribute in the `GradientBoostingMachine` class and slightly change the plotting procedure. Should I do this? Here or another PR?

ogrisel: Plotting the predictor trees is still useful for the users to learn about the structure of the trees. So the plotting utility should still work in that case.

I just want to extend it to make it work with additional info on the grower tree and use it in an example folder stating explicitly that this is only to analyze the performance profile of a single small tree. E.g.:

```
example/plot_performance_profile_single_small_tree.py
```

pygbm/grower.py
```diff
@@ -11,12 +12,24 @@ class TreeNode:
     left_child = None  # Link to left node (only for non-leaf nodes)
     right_child = None  # Link to right node (only for non-leaf nodes)
     value = None  # Prediction value (only for leaf nodes)
+    histograms = None  # array of histogram shape = (n_features, n_bins)
+    sibling = None  # Link to sibling node, None for root
+    parent = None  # Link to parent node, None for root
+    # Computation time of the histograms, or more precisely time to compute
+    # splitability, which may involve some useless computations
+    time = 0
+    ratio = 1  # sibling.time / node.time if node.hist_subtraction, else 1
```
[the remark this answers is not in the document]
NicolasHug: do you mean `time / sample_indices.shape[0]` ?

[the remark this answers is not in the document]
ogrisel: Processing speed is `sample_indices.shape[0] / time` in samples per seconds.

pygbm/grower.py
NicolasHug: Got it

There's a lot of

```python
tic = time()
blablah()
toc = time()
time_spent = toc - tic
```

Do we want a context manager that would allow us to do something like

```python
with Timer() as time_spent:
    blahblah()
```

?

ogrisel: As you wish but I think it's fine like it is now. Using a context manager will add an indentation level.

NicolasHug: Ok let's keep it this way then

ogrisel:
Ok merged!

ogrisel:
I rebased / merged instead of squash merging... Too bad for the history of this project...

ogrisel:
BTW I also pushed 3f7c36fd4d18cb15fdc4cc5623801bd4f45ababd to fix the plotting example timings (pre-compiling). The results are now constant: the speed is always faster when using histogram substraction.

PR #15, opened 2018-10-11
Disable workaround for numba threading backend fix

ogrisel:
The wheels for numba 0.40.1 now use tbb by default without causing the openmp backend introspection to crash (openmp is disabled in those wheels).


ogrisel:
Code coverage report is not very clever to say the least: how can deleting 2 lines of code decrease coverage...

PR #16, opened 2018-10-22
Improvements in pygbm.splitting

ogrisel:
Remove some unnecessary casts in by typing `constant_hessian_value` explicitly and change how arguments are passed internally in pygbm.splitting.

The result is neutral from a performance point of view but avoids compiling useless variants with float64 argument in `_split_gain` for instance.


ogrisel:
@NicolasHug let me know if you have any comment on this small refactoring.

Nitpick: note that I tried to follow PEP 257 for the docstrings: oneliner short summary / blank line / more details. This is important to make IDEs and API documentation tools like sphinx work better.

NicolasHug:
Cool, looks better!

Ok for the docstring I'll be more careful ^^

PR #18, opened 2018-10-29
Added test to compare with LGBM predictions

NicolasHug:
Added a test to compare with LGBM predictions for easy targets with pre-binned data.

I've added the lightgbm dependency for travis to run it but maybe we need a `requirements_travis.txt` file?


tests/test_compare_lightgbm.py
```diff
@@ -0,0 +1,50 @@
+try:
+    import lightgbm as lb
+except ImportError:
+    exit(0)
```
ogrisel: Instead of doing this, please use:

```python
lb = pytest.importorskip("lightgbm")
```

at the beginning of each of the tests that use lb.

tests/test_compare_lightgbm.py
```diff
@@ -0,0 +1,50 @@
...
+@pytest.mark.parametrize('n_samples', [255, 1000])
+def test_same_predictions_easy_target(seed, n_samples):
+    # Make sure pygbm has the same predictions as LGBM for very easy targets.
+    #
+    # Notes:
+    # - As some splits maye have equal gains (and because of float errors) when
+    #   the number of samples in a node is low, it makes more sense to compare
+    #   the predictions rather than the trees which may be split in different
+    #   features just out of luck.
+    # - To avoid discrepancies in the binning strategy, data is pre-binnes if
```
ogrisel: binnes => binned.

tests/test_compare_lightgbm.py
```diff
@@ -0,0 +1,50 @@
...
+                                    learning_rate=1,
+                                    min_child_samples=min_sample_leaf)
+    est_pygbm = GradientBoostingMachine(max_iter=max_iter,
+                                        validation_split=None, scoring=None,
+                                        min_samples_leaf=min_sample_leaf)
+    est_lightgbm.fit(X, y)
+    est_pygbm.fit(X, y)
+
+    pred_lgbm = est_lightgbm.predict(X)
+    pred_pygbm = est_pygbm.predict(X)
```
ogrisel: Can you split X and y using `train_test_split` and then compare predictions both on the training samples and hold out test samples?

NicolasHug:
Thanks, for test data, I make sure that more than 80% of the preditions are equal up to the 5th decimal. We can't really use `assert_array_almost_equal` anymore because some samples may land in different leaves with completely different values.

Alternatively I could compute the MSE?

ogrisel:
For the case when `n_samples = 1000` they should be exactly the same trees if you limit the number of leaves to a small value such as 8 for instance. Maybe test equal predictions on the tests only in that case.

NicolasHug:
Ok, I chose 6 and remove the seed `2` because it would fail (for a single example)

ogrisel:
I pushed a change to only run the comparison on the test set in the identifiable cases (large `n_samples`, low `max_leaf_nodes`) while still comparing predictions on the training set for fully developed trees with a small number of samples.

ogrisel:
I don't need to select the random seed any more with this testing strategy.

ogrisel:
Merged! Thanks @NicolasHug!

PR #19, opened 2018-10-29
[MRG] Parallelized split_indices()

NicolasHug:
I implemented something similar to [what lightgbm is doing](https://github.com/Microsoft/LightGBM/blob/master/src/treelearner/data_partition.hpp#L109).

The splitting time with higgs dataset goes from 3s to 2s so it's not a huge improvement but still worth it I guess. I think it could be further improved by avoiding to allocate all those arrays (each of size n_samples) for each split, and instead allocate them once and for all in the grower. (unrelated, but the same goes for `ordered_gradients` and `ordered_hessians`)

Pushing it as-is, but maybe it'd be best to convert the method into a function, and directly use the `@njit(parallel=True)` decorator which would avoid the 2 sub functions.


ogrisel:
On an oldish machine with 12 physical cores, this PR make the Higgs boson benchmark go from 17-18s down to 14-15s while LightGBM is in the 10-12s range.

Note that the compilation time is also in the 15s range (but could be cached if needed)...

ogrisel:
Can you please rebase this on top of master to check that it does not break the newly merged lightgbm comparison tests?

pygbm/splitting.py
```diff
@@ -60,78 +61,163 @@ def __init__(self, n_features, binned_features, n_bins,
...
-        sum_gradients = ordered_gradients.sum()
-        if self.constant_hessian:
-            n_samples = sample_indices.shape[0]
-            sum_hessians = self.constant_hessian_value * float32(n_samples)
-        else:
-            sum_hessians = ordered_hessians.sum()
+        start, stop = starts[thread_i], starts[thread_i] + left_cnt[thread_i]
+        for i, sample_idx in enumerate(range(start, stop)):
+            sample_indices_left[left_write_offset[thread_i] + i] = \
+                tmp_left[sample_idx]
```
ogrisel: The `sample_idx` variable name here is confusing. This integer is not an index of a sample in the original training set feature array. It could be renamed to something more neutral such as j.

Alternatively, I wonder if numba accepts something like:

```python
sample_indices_left[left_write_offset[thread_i]:left_write_offset[thread_i]+left_cnt[thread_i]] = tmp_left[start:stop]
```

Off-course the above is ugly and should be rewritten with additional temporary index variables to make it more readable but you get the idea.

pygbm/splitting.py
```diff
@@ -60,78 +61,163 @@ def __init__(self, n_features, binned_features, n_bins,
...
+    #   help of left_write_offset and right_write_offset, whose entries are
+    #   here denoted by '^':
+    #   - sample_indices_left =   [abefilmnopqrtux.........]
+    #                              ^   ^ ^     ^
+    #   - samples_indices_right = [cdghjksvw...............]
+    #                              ^ ^   ^
+    #                                    ^
+
+    # Note: we could get rid of the tmp_xxxx arrays in our case and
+    # directly work on samples_indices_left/right. I'm leaving them
```
ogrisel: Have you tried to use samples_indices_left / samples_indices_right directly instead of using the `tmp_*` arrays? I suspect this would run faster as you don't need to copy the indices at the end of this function (the last `prange` loop).

pygbm/splitting.py
ogrisel: In the future, if we want to reuse the `sample_indices_left`/`sample_indices_right` buffers for consecutive splits in the same tree we could always change the returned values to return copies to be stored on the nodes that result from the split:

```python
    return (context.sample_indices_left[:total_left_count].copy(),
               context.sample_indices_right[:total_right_count].copy())
```

pygbm/splitting.py
NicolasHug: I can try but I don't think we'll gain much, the second prange loop is still needed because there is no way to know in advance how many samples land into the left or right node, for each thread. So we would gain the array allocation time only.

pygbm/splitting.py
NicolasHug: > we could always change the returned values to return copies

Yes, but that wouldn't save any time / memory right? The copied array would still need to be allocated.

Let me try to explain what LightGBM is doing to avoid those allocations at every node:

To keep track of how the samples are dispatched over the leaves of the tree, Lightgbm uses a `DataPartition` object. There's a one-to-one relationship between a grower and a data_partition. The main structure of this `DataPartition` is a `indices_` array which looks like this:

`[qwert|yuio|pa|sdfghjk|lzxc]`

which basically indicates that the samples `qwert` are in the first leaf, `yuio` in the second, etc.

If the first leaf is split, the array would become, e.g.

`[qt|erw|yuio|pa|sdfghjk|lzxc]` (the order of the samples inside a leaf is irrelevant)
This is why they need the `tmp...` array in the `split_indices` method BTW, they can't work directly on `indices_` because they would overwrite some parts of it.

pygbm/splitting.py
ogrisel: Indeed. Let's keep this way then. But remove this comment then.

pygbm/splitting.py
NicolasHug: For slices:

Tests pass with `parallel=False` but not with `parallel=True`. Not sure if this is a numba bug because I could not reproduce it. I'll keep the current version for now.

```python
        start_tmp = starts[thread_idx]
        stop_tmp = starts[thread_idx] + left_counts[thread_idx]
        start = left_write_offset[thread_idx]
        stop = left_write_offset[thread_idx] + left_counts[thread_idx]

        sample_indices_left[start:stop] = tmp_left[start_tmp:stop_tmp]
```

pygbm/splitting.py
```diff
@@ -60,78 +61,160 @@ def __init__(self, n_features, binned_features, n_bins,
...
+        start = starts[thread_idx]
+        stop = starts[thread_idx] + right_counts[thread_idx]
+        for i, j in enumerate(range(start, stop)):
+            sample_indices_right[right_write_offset[thread_idx] + i] = \
+                tmp_right[j]
+
+    total_left_cnt = left_counts.sum()
+    total_right_cnt = right_counts.sum()
+    return (sample_indices_left[:total_left_cnt],
+            sample_indices_right[:total_right_cnt])
```
ogrisel: Could be simplified to:

```python
    return (sample_indices_left[:left_counts.sum()],
            sample_indices_right[: right_counts.sum()])
```

pygbm/splitting.py
ogrisel: Alright. Thanks for the explanation.

ogrisel:
Thanks, merged!

PR #21, opened 2018-10-31
[MRG] Avoid multiple array allocations by using a partition array

NicolasHug:
This PR avoids the allocations of the `samples_indices` array for each nodes.

Indexes are stored in the `partition` array in `SplittingContext` and the `samples_indices` are now views on this array.

Strongly inspired from lightgbm

I'm not sure how better it is performance-wise but it sure doesn't hurt. And memory footprint is now much lower.


NicolasHug:
Marking it as WIP because after having looked at it again, I'll try to directly work on the `sample_indices` views rather than on `partition`. This would avoid storing the nodes positions in the grower which I think is not appropriate.

ogrisel:
I benched the current implementation of this PR and it seems slightly but deterministically slower than the implementation in master (on Higgs boson).

NicolasHug:
Hm ok thanks, I'll keep an eye on this

ogrisel:
The split cost performance with `parallel=True` in this PR is slightly better than on master. Good job.

NicolasHug:
Putting back [MRG] for review.

On average the higgs boson benchmarks runs in 17.5s on my laptop so I don't observe the same slow-down as you did.

When I run the same benchmark on the current master branch, I get about 20s but something weird is going on: it used to be much faster yesterday IIRC and times are highly variable. Some run in 16+s and some in 23...

ogrisel:
The split costs of this PR is very close to the one reported by LightGBM with `TIME_TAG` enabled.

pygbm/splitting.py
```diff
@@ -61,66 +64,96 @@ def __init__(self, n_features, binned_features, n_bins,
...
+        # the indices, e.g.:
+        # partition = [abcdefghijkl]
+        # After a call to split_indices, it may look e.g. like this:
+        # partition = [cef|abdghijkl]
+        # we have 2 leaves, the left one is at position 0 and the second one at
+        # position 3. The order of the samples is irrelevant.
+        self.partition = np.arange(0, binned_features.shape[0], 1, np.uint32)
+        # tmp arrays are used in split_indices to support parallel splitting.
+        self.tmp_left = np.empty_like(self.partition)
+        self.tmp_right = np.empty_like(self.partition)
```
ogrisel: Maybe it's possible to find a more descriptive name for those attributes: `left_indices_buffer`  and `right_indices_buffer`.

ogrisel:
Merged, thanks @NicolasHug!

PR #29, opened 2018-11-01
Fix dependency versions

### LINKED ISSUE #28: scikit-learn 0.20 is required
Please add into your dependencies that scikit-learn 0.20 is required.

Otherwise, it throws the following error:

> ImportError: cannot import name 'check_scoring'

Which is only available from scikit-learn 0.20 (latest stable version).

Laurae2:
Fixes #28 where pygbm requires scikit-learn at least version 0.20 for the function `check_scoring`.

In addition, numba version 0.39 or higher is required for fixing "NotImplementedError: unsupported nested memory-managed object". See https://github.com/numba/numba/pull/2840


ogrisel:
Thanks!

PR #33, opened 2018-11-01
[MRG] Ordered_gradients now in SplittingContext

### LINKED ISSUE #22: Allocate ordered_gradients and ordered_hessians only once
Opening this issue so that we don't forget there is room for improvement on this matter.

We could allocate `ordered_gradients` and `ordered_hessians` once and for all in the `SplittingContext` object instead of doing the allocation for every node.

Also, we could populate them in parallel.

I tried doing this in 1b073b1fc8d5e620f84859177a1021ae231b7fa6. It passes the tests but it still allocates the arrays for each node because I forgot to remove [this line](https://github.com/ogrisel/pygbm/pull/21/commits/1b073b1fc8d5e620f84859177a1021ae231b7fa6#diff-43a3fcc0b6fe7d9cdd257e9b6d2b4fe3L208) (><).

When I did remove it, and after having patched places where `ordered_gradients` is used (e.g. changing `sum_gradients = ordered_gradients.sum()` into`sum_gradients = ordered_gradients[:n_samples].sum()`), the test don't pass anymore. It could be a numba bug similar to https://github.com/numba/numba/issues/3459, but I'm not certain I didn't miss something myself.

NicolasHug:
Closes #22 

This PR:
- Allocates `ordered_gradients` and `ordered_hessians` in `SplittingContext` instead of for each node.
- Puts `sum_gradients` and `sum_hessians` in `SplittingContext` (no performance or memory gain from this, just cleaner code IMHO)
- Parallelizes the population of `ordered_gradients` and `ordered_hessians`

I don't see any clear performance gain on the higgs boson benchmark but it doesn't seem to hurt at all, memory usage will surely be better now.


ogrisel:
I like the approach. I think the memory benefit will only be visible when recycle the sample `TreeGrower` instance (and therefore the same `SplittingContext` instance) over and over for all the trees.

pygbm/splitting.py
```diff
@@ -192,38 +201,48 @@ def split_indices(context, split_info, sample_indices):
...
+    if ctx.constant_hessian:
+        ctx.sum_hessians = ctx.constant_hessian_value * float32(n_samples)
     else:
-        sum_hessians = ordered_hessians.sum()
+        ctx.sum_hessians = ctx.ordered_hessians[:n_samples].sum()
 
-    return _parallel_find_split(
-        context, sample_indices, ordered_gradients, ordered_hessians,
-        sum_gradients, sum_hessians)
+    return _parallel_find_split(ctx, sample_indices)
```
ogrisel: Because `_parallel_find_split` is a very short function, I think we could collapse it here. We could reuse the docstring of `_parallel_find_split` directly for the `find_node_split` function that has no docstring.

I believe collapsing the functions together would make the code easier to follow/understand.

pygbm/splitting.py
```diff
@@ -234,19 +253,19 @@ def find_node_split_subtraction(context, sample_indices, parent_histograms,
...
-        sum_hessians = (parent_histograms[0]['sum_hessians'].sum() -
-                        sibling_histograms[0]['sum_hessians'].sum())
+        context.sum_hessians = (parent_histograms[0]['sum_hessians'].sum() -
+                                sibling_histograms[0]['sum_hessians'].sum())
 
-    return _parallel_find_split_subtraction(
-        context, parent_histograms, sibling_histograms,
-        sum_gradients, sum_hessians)
+    return _parallel_find_split_subtraction(context, parent_histograms,
+                                            sibling_histograms)
```
ogrisel: Similar remark as above w.r.t. collapsing function and reusing the docstring. Don't forget to add the `parallel=True` flag to `find_node_split_subtraction` when you do this.

ogrisel:
Merged! Thanks @NicolasHug.

Issue #34, opened 2018-11-02
mean_samples_leaf does not do what it's suppose to do


ogrisel:
`min_samples_leaf=20` should ensure that we never do a split that would result in less than 20 samples in each of the two resulting leaves.

Our current implementation does not split nodes with less than 20 samples which is not the same. Our current implementation is akin to the `min_samples_split` of scikit-learn trees which is not a good hyperparameter to control over-fitting.

I am working on a fix.

ogrisel:
`min_samples_leaf` is now doing what it's supposed to do (I think) but pygbm is still not 100% consistent with LightGBM. So #32 is not resolved yet.

PR #35, opened 2018-11-02
Fix handling of the min_samples_leaf hyperparameter

ogrisel:
This is a tentative fix for #34.

However the test in `test_compare_lightgbm` fails if I change the value of min_samples_leaf to something different than 1.

In retrospect, I believe this is because we reject too many splits by doing the filtering at the grower level while we should probably do it also inside the `find_node_split*` calls.


ogrisel:
I don't have the time to work on this further today. @NicolasHug feel free to takeover if you wish so.

Note that in the current state of this PR the Higgs boson benchmark is significantly slower than LightGBM but this is probably because we reject too many splits by doing the coarse node level filter only (instead of doing per-feature feature filtering as well).

ogrisel:
@NicolasHug I sent you an invite to have commit rights to this repo.

This PR needs a rebase on top of master to fix the conflicts.

NicolasHug:
Ok I'll take it up

>we reject too many splits by doing the filtering at the grower level while we should probably do it also inside the find_node_split* calls.

Yes I think this should be at the histogram splitting level [like in lightgbm](https://github.com/Microsoft/LightGBM/blob/master/src/treelearner/feature_histogram.hpp#L533).


Do we want to rename the current `min_samples_leaf` into `min_samples_split` to be more scikit-learnesque, or completely get rid of it?

ogrisel:
No let's get rid of the old `min_samples_split` strategy to keep the code as simple as possible. It's was not a good way to control for overfitting. `min_samples_leaf` is a better strategy.

pygbm/splitting.py
```diff
@@ -307,12 +311,12 @@ def _parallel_find_split_subtraction(context, parent_histograms,
     histograms by substraction.
     """
     # Pre-allocate the results datastructure to be able to use prange
-    split_infos = [SplitInfo(0, 0, 0, 0., 0., 0., 0.)
+    split_infos = [SplitInfo(0, 0, 0, 0., 0., 0., 0., 0, 0)
                    for i in range(context.n_features)]
```
ogrisel: This data structure could probably be also stored as an attribute on the context to avoid reallocating it over and over again.

pygbm/splitting.py
```diff
@@ -362,26 +368,29 @@ def _find_histogram_split_subtraction(context, feature_idx,
...
+        n_samples)
 
 
-@njit(locals={'gradient_left': float32, 'hessian_left': float32},
+@njit(locals={'gradient_left': float32, 'hessian_left': float32,
+              'n_samples_left': uint32},
       fastmath=True)
 def _find_best_bin_to_split_helper(context, feature_idx, histogram,
-                                   sum_gradients, sum_hessians):
+                                   sum_gradients, sum_hessians, n_samples):
```
ogrisel: I had not realized but it's probably stupid to pass n_samples as an argument here. The information is already passed as part of the context.

NicolasHug: I don't think it's in the context, I thought about it but all the arrays are of size `n_samples` which is the size of the whole training set

ogrisel: Yes, this is what I meant.

ogrisel: You are right.

NicolasHug:
I did the following changes:

- `min_samples_leaf` is now checked at the histogram level, like in lightgbm.
- `min_gain_to_split` as well.
- had to set `decimal=3` in `test_compare_lightgbm`. Like for you, tests don't pass for `mean_samples_leaf > 1`, but the predictions are still pretty close in general.

- it doesn't run slower than on master. I also observed that your last commit was a lot slower and I don't understand why: less splits to consider = less work = faster, as far as I understand. Maybe some numba compilation thing?

- tests fail on `test_predictor` because I set `mean_samples_leaf` to 5 and slightly modified the code to plot the lightgbm model. I get this (lightgbm tree is on the left):
[Digraph.gv.pdf](https://github.com/ogrisel/pygbm/files/2543945/Digraph.gv.pdf). I don't understand why our tree is going so deep on the same feature.

- I get the exact same ROC AUC as on master: .7892

Things are not totally broken there's something fishy going on.

NicolasHug:
The code from `test_predictor` ran with master gives a much more reasonable tree: [Digraph.gv.pdf](https://github.com/ogrisel/pygbm/files/2543964/Digraph.gv.pdf)

pygbm/splitting.py
```diff
@@ -76,6 +78,7 @@ def __init__(self, n_features, binned_features, n_bins,
         self.l2_regularization = l2_regularization
         self.min_hessian_to_split = min_hessian_to_split
         self.min_samples_leaf = min_samples_leaf
+        self.min_gain_to_split = min_gain_to_split
```
ogrisel: You should raise a `ValueError` if the user passes `min_gain_to_split < 0`.

tests/test_predictor.py
```diff
@@ -20,9 +20,24 @@ def test_boston_dataset():
...
+    lb = pytest.importorskip("lightgbm")
+    est_lightgbm = lb.LGBMRegressor(n_estimators=1,
+                                    min_data_in_bin=1,
+                                    learning_rate=1,
+                                    min_data_in_leaf=min_samples_leaf,
+                                    num_leaves=max_leaf_nodes)
+    est_lightgbm.fit(X_train_binned, y_train)
+
+    from pygbm.plotting import plot_tree
+    plot_tree(grower, est_lightgbm)
```
ogrisel: Please do not put such plots in the regular tests. Use the `examples/` folder for visual debugging instead.

ogrisel: Also if you want to compare lightgbm and pygbm on the Boston dataset, please add a new test in `test_compare_lightgbm.py` instead.

ogrisel:
Indeed there is something fishy going on...

ogrisel:
test_predictor.py used to pass before 7e91c2c454691faffc2192e75992c995a2792d3c, right? Maybe this commit is causing the regression but I am not sure why.

pygbm/splitting.py
```diff
@@ -364,45 +378,62 @@ def _find_histogram_split_subtraction(context, feature_idx,
...
+            hessian_left += (histogram[bin_idx]['count']
+                             * context.constant_hessian_value)
         else:
             hessian_left += histogram[bin_idx]['sum_hessians']
         if hessian_left < context.min_hessian_to_split:
             continue
         hessian_right = context.sum_hessians - hessian_left
         if hessian_right < context.min_hessian_to_split:
-            continue
+            # won't get any better
```
ogrisel: Maybe add a comment to say that the loss functions are all convex and therefore the hessians are positive.

pygbm/splitting.py
```diff
@@ -364,45 +378,62 @@ def _find_histogram_split_subtraction(context, feature_idx,
...
+
+        gradient_left += histogram[bin_idx]['sum_gradients']
         gradient_right = context.sum_gradients - gradient_left
         gain = _split_gain(gradient_left, hessian_left,
                            gradient_right, hessian_right,
                            context.sum_gradients, context.sum_hessians,
                            context.l2_regularization)
-        if gain > best_split.gain:
+
+        if gain > best_split.gain and gain > context.min_gain_to_split:
```
ogrisel: We should probably have `gain >= context.min_gain_to_split`.

ogrisel: Yes this is the cause of the bad trees.

ogrisel: Oops I spoke too quickly, I had set `min_samples_leaf=1` in `test_predictor.py` and forgot about it.

tests/test_predictor.py
NicolasHug: Sure, this is not meant to stay, I just left it so you can reproduce my plots if you wanted to

pygbm/grower.py
```diff
@@ -213,16 +210,27 @@ def split_next(self):
...
+                and left_child_node.n_samples < self.min_samples_leaf * 2):
+            self._finalize_leaf(left_child_node)
         else:
             self._compute_spittability(left_child_node)
+
+        if (self.min_samples_leaf is not None
+                and right_child_node.n_samples < self.min_samples_leaf * 2):
+            self._finalize_leaf(right_child_node)
+        else:
             self._compute_spittability(right_child_node)
```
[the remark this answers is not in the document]
NicolasHug: lightgbm ^^

ogrisel:
Maybe the structure we observe is expected in cases the most predictive feature has a value that is linearly correlated with the target value: the gain should be constant for consecutive `bin_idx` for a given feature. One way to mitigate this would be to detect those areas of constant gains and split in the middle of the plateau instead of the right hand-side.

ogrisel:
I am not sure about my theory, I cannot come up with a synthetic dataset that would trigger this case.

pygbm/splitting.py
NicolasHug: There's already a check in the grower. BTW `min_gain_to_split` is only a parameter of the grower, not of `GradientBoostingMachine`. We should move it there right?

NicolasHug:
> test_predictor.py used to pass before 7e91c2c, right? Maybe this commit is causing the regression but I am not sure why.

Yes it passes in 1c6e62b with 5 leaves. The threshold was still lowered from .9 to .75.

pygbm/splitting.py
ogrisel: Right this is fine for now. No need to expose it to the public API at this time.

ogrisel:
> Yes it passes in 1c6e62b with 5 leaves. The threshold was still lowered from .9 to .75.

`min_samples_leaf=5`, not 5 leaves.

The threshold on the training set is expected to drop when we control overfitting with a stricter `min_samples_leaf`. However there should be a value where the test accuracy was comparable to the previous values. I just pushed one such value.

Issue #31, opened 2018-11-01
Excessive memory usage when the number of trees is large (e.g. 500 trees)


ogrisel:
As investigated in #30, when the number of trees is large, htop reports VIRT memory that exceeds the physical memory of the host and causes computer hangs that slows down computation.

What is weird is that RES memory seems to stay constant.

ogrisel:
This problem can be reproduces even on very small training sets (e.g. 1e4 samples) but with `n_trees > 500`.

NicolasHug:
This should hopefully be fixed by #21 right? At least partially

ogrisel:
> This should hopefully be fixed by #21 right? At least partially

Memory use is still much higher with pygbm. Maybe we could try to reuse the grower instance over and over again with some kind of a  `.reset(gradients, hessians)` method.

NicolasHug:
>with some kind of a .reset(gradients, hessians) method.

that's exactly what LGBM is doing BTW. I think it would make sense yes, 500 trees = 500 array allocations of size n_samples so it takes a toll.

But that means somehow the old grower instances aren't properly garbage collected?

ogrisel:
BTW, the split costs on master is still 2x faster than on this PR.

ogrisel:
> But that means somehow the old grower instances aren't properly garbage collected?

This is what I suspect. Maybe numba is keeping some internal references on jitclass instances?

NicolasHug:
So that would suggest that the problem doesn't come from some garbage collector issue?

I tried a quick hack

 ```python

    # reset grower
    def reset(self, all_gradients, all_hessians, shrinkage=1.):
        self.splitting_context.reset(all_gradients, all_hessians)
        self.shrinkage = shrinkage
        del self.splittable_nodes  # just to make sure
        del self.finalized_leaves
        self.splittable_nodes = []
        self.finalized_leaves = []
        self.total_find_split_time = 0.  # time spent finding the best splits
        self.total_apply_split_time = 0.  # time spent splitting nodes
        self._intilialize_root()
        self.n_nodes = 1
```

```python
    # reset split info
    def reset(self, all_gradients, all_hessians):
        self.all_gradients[:] = all_gradients
        self.all_hessians[:] = all_hessians
        self.sum_gradients = self.all_gradients.sum()
        self.sum_hessians = self.all_hessians.sum()
        self.ordered_gradients[:] = all_gradients.copy()
        self.ordered_hessians[:] = all_hessians.copy()
```

[...]

PR #36, opened 2018-11-05
[WIP] Do not store histogram on SplitInfo

ogrisel:
Tentative fix for #31.


ogrisel:
At this time some tests are broken (need to be updated) and the code runs slightly slower than on master for some reason I do not understand yet.

The memory leak issue should be fixed though.

ogrisel:
The fact that we observe a slowdown when we update the packed `histograms` array in the parallel for loop while we do not observe this issue in master where the packed histograms array is filled sequentially might be a case of [False Sharing](https://en.wikipedia.org/wiki/False_sharing).

ogrisel:
Actually I tried again with master and the various threading backends and this PR is either as fast or faster than master. I must have done something wrong when I reported the initial slow down.

In any case, tbb is significantly faster than the workqueue backend when the number of cores is large (e.g. 12 in my case).

ogrisel:
@NicolasHug I have to go, feel free to update the failing tests and merge this PR.

ogrisel:
Cool, let's merge as it's already a net improvement.

About the minimal reproduction case, I confirm I get the leak with your code, without any error message or exception. Just memory usage increasing as reported by psutil.

ogrisel:
We still have a discrepancy in terms of results with LightGBM though. But there is another issue for that.

NicolasHug:
I have opened https://github.com/numba/numba/issues/3473 and https://github.com/numba/numba/issues/3472 regarding the leak and some other weird stuff I found.

PR #37, opened 2018-11-05
[MRG] Added n-trees and n-leaf-nodes as script parameters to bench_higgs_boson

NicolasHug:


ogrisel:
Thanks, I pushed even more options in my last commit.

Issue #32, opened 2018-11-01
Investigate the discrepancy in default hyperparams compared to LightGBM


ogrisel:
Possible culprits:

- `shrinkage` / `learning_rate`
- `min_samples_leaf`
- `min_child_samples`

See details in https://github.com/ogrisel/pygbm/issues/30#issuecomment-435091127.

guolinke:
@NicolasHug I think you used the wrong parameter name in that code.

ogrisel:
Indeed. It's actually the pygbm handling of `min_samples_leaf` that is broken. See: #34.

PR #40, opened 2018-11-07
[MGR] Strict comparison to min_gain_to_split

NicolasHug:
I think the comparison to `min_gain_to_split` should be a strict comparison.

With the current non-strict comparison and with e.g. `min_gain_to_split=0`, a tree will keep on growing even if all the leaves are completely pure.


ogrisel:
Could you please add a test for this change? Also, the code in the grower should be updated to be consistent:

```diff
diff --git a/pygbm/grower.py b/pygbm/grower.py
index ea77473..b8fb977 100644
--- a/pygbm/grower.py
+++ b/pygbm/grower.py
@@ -159,7 +159,7 @@ class TreeGrower:
             # compute the histogram.
             return
 
-        if node.split_info.gain < self.min_gain_to_split:
+        if node.split_info.gain <= self.min_gain_to_split:
             # Note: this also comprises the case where there is not enough
             # samples in left or right child: all the SplitInfo (including the
             # best one which is returned by find_node_split_xxx) would have a

```

There might be other places where the code needs to updated as well.

ogrisel:
I would rather like to never use `==` on a float32 value :)

So maybe `if node.split_info.gain < 0`.

tests/test_compare_lightgbm.py
```diff
@@ -59,9 +59,9 @@ def test_same_predictions_easy_target(seed, n_samples, max_leaf_nodes):
 
     pred_lgbm = est_lightgbm.predict(X_train)
     pred_pygbm = est_pygbm.predict(X_train)
-    np.testing.assert_array_almost_equal(pred_lgbm, pred_pygbm, decimal=5)
+    np.testing.assert_array_almost_equal(pred_lgbm, pred_pygbm, decimal=3)
 
     if max_leaf_nodes < 10 and n_samples > 1000:
         pred_lgbm = est_lightgbm.predict(X_test)
         pred_pygbm = est_pygbm.predict(X_test)
-        np.testing.assert_array_almost_equal(pred_lgbm, pred_pygbm, decimal=5)
```
ogrisel: I don't like degrading the strictness of this test without understanding what's going on first. 

What is the behavior of LightGBM in the case where all leaves are pure (zero gain) before reaching the `max_leaf_nodes` limit?

NicolasHug: Yes I don't like it either.

The test fails for only one seed, on those 2 samples:

```python
    cond = abs(pred_lgbm - pred_pygbm) > 1e-5
    print(pred_lgbm[cond])
    print(pred_pygbm[cond])
```

```
[2.59808545 2.59740119]
[2.5977433 2.5977433]
```

My intention was to lower the threshold now for the tests to pass (1e-3 is still decent IMHO) and go back to it later when we address #32 fully.

>What is the behavior of LightGBM in the case where all leaves are pure (zero gain) before reaching the max_leaf_nodes limit?

They [stop growing the tree](https://github.com/Microsoft/LightGBM/blob/master/src/treelearner/serial_tree_learner.cpp#L207), as far as I understand (just like we are).

ogrisel: Alright.

ogrisel:
Merged!

PR #42, opened 2018-11-20
[MRG] Fix handling of number of bins

NicolasHug:
The current code does not correctly handle the number of bins in [this loop](https://github.com/ogrisel/pygbm/blob/master/pygbm/splitting.py#L400) because `context.n_bins` is always equal to the `max_bins` parameter [passed in GradientBoostingMachine](https://github.com/ogrisel/pygbm/blob/master/pygbm/gradient_boosting.py#L102).

`n_bins` should be dependent on the feature, and it should also reflect the fact that when `max_bins > n_unique_values`, as we use the unique values as thresholds, the actual number of bins will be equal to `n_unique_values - 1 < max_bins`. The `-1` comes from the fact that we don't actually take the unique values, but the midpoints.

I think this is (one of?) the reason why we're observing overflow in the hessians computations with logloss in #13 .


pygbm/grower.py
```diff
@@ -70,8 +71,13 @@ def __init__(self, features_data, all_gradients, all_hessians,
         if not features_data.flags.f_contiguous:
             warnings.warn("Binned data should be passed as Fortran contiguous"
                           "array for maximum efficiency.")
+        if isinstance(n_bins_per_feature, int):
+            n_bins_per_feature = np.array(
+                [n_bins_per_feature] * features_data.shape[1],
+                dtype=np.uint32)
```
ogrisel: I would pass `n_bins_per_features=None` and `max_bins=256` to the constructor and then:

```python
if n_bins_per_features is None:
    n_bins_per_features = max_bins

if isinstance(n_bins_per_feature, int):
    n_bins_per_feature = np.array(
        [n_bins_per_feature] * features_data.shape[1],
        dtype=np.uint32)
```

NicolasHug:
Thanks for the comments! I actually made a mistake in the first commit: `n_bins = n_thresholds + 1` of course.

I've changed the default value of `max_bins` to 256 (it was already 256 for `BinMapper` but not for `GradientBoostingMachine` nor `TreeGrower`). Note that LightGBM's  default is 255, but maybe that's because they artificially add some thresholds and still end up with 256 bins in the end.

The test in [test_predictor](https://github.com/ogrisel/pygbm/pull/42/files#diff-f7ed36be9d6b8b0e2579011fc9929714R14) fails for anything other than `max_bins=255`. I'm not sure why but that doesn't seem right to me. This is already the case on the master branch BTW.

pygbm/splitting.py
```diff
@@ -349,19 +353,19 @@ def _find_histogram_split(context, feature_idx, sample_indices):
     if root_node:
         if context.constant_hessian:
             histogram = _build_histogram_root_no_hessian(
-                context.n_bins, binned_feature, ordered_gradients)
+                context.max_bins, binned_feature, ordered_gradients)
```
NicolasHug: Actually we should also pass `n_bins_per_feature[feature_idx]` to all histogram computation routines.

`max_bins` is used to create histograms with a uniform shape, and `n_bins_per_feature[feature_idx]` is used to iterate over the relevant bins only

ogrisel:
Merged. Please open an issue to track the test that fails with `max_bins != 255`.

PR #43, opened 2018-11-23
[MRG] Fixed computation of hessian/gradient left/right

NicolasHug:
There was a bug in the most inner splitting loop: if we hit a `continue` because of `min_samples_leaf`, the gradients and hessians values would not be correctly updated in the subsequent bins.

This fixes the discrepancies in `test_compare_lightgbm`, and also some of the overflow issue #13 (but not all of them)


ogrisel:
Good catch. The code is also much more natural in that order (in addition to be correct :)

PR #46, opened 2018-11-23
[MRG] min_samples_leaf now only accepts int values (not None)

NicolasHug:
`min_samples_leaf` used to accept `None` but IMHO it only makes the code slightly more complicated. `min_samples_leaf=1` seems just as natural to me so this PR forces to pass an int. We don't need `numba.optional` anymore.


ogrisel:
Good!

PR #49, opened 2018-11-30
[MRG] Add support for multiclass classification

NicolasHug:
Looks like it's working, still needs lots of cleaning and testing.


NicolasHug:
This is now in a decent shape I think. Putting as MRG if you want to have a look, but we should probably merge #13 first (I branched off this)

ogrisel:
I messed up with git and corrupted my stashed changes when trying to update them w.r.t. you last commit. I have to go to the airport soon. I might be able to connect back from there in two hours . Feel free to go on tweaking in the mean time if you wish otherwise I will try to get back to it from there.

pygbm/loss.py
```diff
@@ -104,5 +114,82 @@ def _update_gradients_hessians_logistic(gradients, hessians, y_true, y_pred):
...
+            for i in range(starts[thread_idx], ends[thread_idx]):
+                # p_k is the probability that class(ith sample) == k.
+                # This is a regular softmax. Ideally we should be using
+                # logaddexp or logsumexp but they are not supported by numba.
+                denominator = np.sum(np.exp(y_pred[i, :]))
+                p_k = np.exp(y_pred[i, k] - np.log(denominator))
+                gradients_at_k[i] = p_k - (y_true[i] == k)
+                hessians_at_k[i] = p_k * (1. - p_k)
+                # LightGBM uses 2 * p_k * (1 - p_k) but that seems incorrect?
+                # Looks like it gives much better results in practice though.
```
NicolasHug: @ogrisel This is ready for reviews, but this discrepancy in particular is confusing to me

ogrisel: Have you tried to compare with different values of the learning rate? Doubling the hessian is equivalent to decreasing the learning rate by 2, isn't it?

pygbm/loss.py
```diff
@@ -104,5 +114,82 @@ def _update_gradients_hessians_logistic(gradients, hessians, y_true, y_pred):
...
+                # Both versions pass the compare_lgbm tests but that's because
+                # we only build 1 tree.
+                # lightgbm version does not pass the numerical hessian test.
+
+                # Note: I observed super weird numba bugs here.
+                # This fails:
+                # p_k = np.exp(y_pred[i, k] - np.log(np.sum(np.exp(y_pred[i]))
+                # This fails as well
+                # p_k = np.exp(y_pred[i, k] -
+                #              np.log(np.sum(np.exp(y_pred[i, :]))
```
ogrisel: Shouldn't we use a numerically stable primitive such as the code in `scipy.special.logsumexp`?

https://github.com/scipy/scipy/blob/master/scipy/special/_logsumexp.py#L9-L129

This is not a primitive in numpy but the code is pure numpy so numba should be able to optimize it in nopython mode although I have not checked. If that's not the case we will have to maintain our own numba-compatible implementation of logsumexp in pygbm.

pygbm/multiclass_notes
```diff
@@ -0,0 +1,45 @@
...
+as a regular softmax between scores = [scores_0, scores_1, ... scores_K-1]
+where scores_k = sum(<leaf value of kth tree of iteration i>
+                     for i in range(n_iterations)).
+The predicted class is then the argmax of the K probabilities.
+
+Regarding implementation details, the arrays
+- y_pred
+- gradients
+- hessians (for non-constant hessians)
+- decision, as returned by _decision_function()
```
ogrisel: This doc should be updated to stop mentionning `y_pred` and `_decision_function` and instead use `raw_predictions` and `_raw_predict`.

pygbm/loss.py
NicolasHug: I made a comment about this a few lines above in the code.

The function we need is `scipy.logsumexp` but it's not supported. `np.logaddexp.reduce` is equivalent, but numba does not support `reduce` it seems.


```python
from numba import njit
import numpy as np

@njit
def g():
    a = np.arange(10).astype(np.float32)
    b = np.logaddexp.reduce(a)
    return b

print(g())

```

```
numba.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Unknown attribute 'reduce' of type Function(<ufunc 'logaddexp'>)

```

pygbm/loss.py
NicolasHug: You are right, dividing the learning rate by 2 should gives the same results (as well as `min_hessian_to_split` but that's not critical here). And it does give the same results, provided that we change 

`shrinkage = 1. if self.n_iter_ == 0 else self.learning_rate`

to just

`shrinkage = self.learning_rate`.

Setting the shrinkage it to 1 for the first iteration seems to hurt accuracy quite a bit, and [LightGBM](https://github.com/Microsoft/LightGBM/blob/master/src/boosting/gbdt.cpp#L371) applies shrinkage at every iteration including the first one.

Should we change this?

pygbm/loss.py
ogrisel: But the code of logsumexp is only doing basic numpy operation so it should work in numba nopython mode, no? If not we could / should adapt our own copy of it in pygbm to make it run in `@njit` mode

NicolasHug: I'm not sure that numba supports functions defined in scipy. We had the same issue with `expit`.


```
numba.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Untyped global name 'logsumexp': cannot determine Numba type of <class 'function'>

File "numba_lol.py", line 8:
def g():
    <source elided>
    a = np.arange(10).astype(np.float32)
    b = logsumexp(a)
    ^
```

You mean we should have our own `logaddexp` **with** numerical stability?
Because we already have the basic (potentially unstable) version here in the code

NicolasHug:
How about we change the `all_gradients` and `all_hessians` parameters of `TreeGrower` to just `gradients` and `hessians`?

I think it is a bit ambiguous because in a multiclass context, they are not "all the gradients" anymore, just those of the kth tree.

pygbm/loss.py
ogrisel: You need to wrap it with the `njit` decorator before calling it from an `njit`ed function.

pygbm/loss.py
ogrisel: Yes. I did that because I thought that LightGBM was doing but maybe I misread the LightGBM code.

pygbm/loss.py
NicolasHug: Like this you mean? Still getting the error (all functions are njitted even though it doesn't show on the trace)

```
Untyped global name 'logsumexp': cannot determine Numba type of <class 'function'>

File "numba_lol.py", line 7:
def _logsumexp(x):
    return logsumexp(x)
    ^

[1] During: resolving callee type: type(CPUDispatcher(<function _logsumexp at 0x7f761c3951e0>))
[2] During: typing of call at numba_lol.py (12)


File "numba_lol.py", line 12:
def g():
    <source elided>
    a = np.arange(10).astype(np.float32)
    b = _logsumexp(a)
```

The numerical stability trick for logsumexp is [a classic](http://cs231n.github.io/linear-classify/#softmax) I think, I just pushed something.

NicolasHug:
I think I addressed all comments so far @ogrisel :)

pygbm/gradient_boosting.py
```diff
@@ -138,62 +142,110 @@ def fit(self, X, y):
...
+            shape=(n_samples * self.n_trees_per_iteration), dtype=np.float32)
+        # Should we parallelize this?
+        for predictors_of_ith_iteration in self.predictors_:
+            for k, predictor in enumerate(predictors_of_ith_iteration):
+                start, end = n_samples * k, n_samples * (k + 1)
+                raw_predictions[start:end] += predictor.predict(X)
 
         return raw_predictions
 
     def _predict_binned(self, X_binned):
```
ogrisel: Maybe this should also be renamed `_raw_predict_binned`.

pygbm/loss.py
```diff
@@ -1,11 +1,27 @@
...
 import numba
 
 
+# TODO: Write proper docstrings
+
+@njit
+def _logsumexp(a):
+    # custom logsumexp function with numerical stability, based on scipy's
+    # logsumexp which is unfortunately not supported (neither is
+    # np.logaddexp.reduce). Only supports with 1d arrays.
```
ogrisel: typo: Only supports 1d arrays.

pygbm/loss.py
```diff
@@ -104,5 +129,67 @@ def _update_gradients_hessians_logistic(gradients, hessians, y_true, y_pred):
...
+        one_hot_true = np.zeros_like(y_pred)
+        for k in range(self.n_trees_per_iteration):
+            one_hot_true[:, k] = (y_true == k)
+
+        return (logsumexp(y_pred, axis=1) -
+                (one_hot_true * y_pred).sum(axis=1))
+
+    def update_gradients_and_hessians(self, gradients, hessians, y_true,
+                                      y_pred, k=0):
+        # reshape will hopefully return a view so it's cheap?
```
ogrisel: yes

pygbm/multiclass_notes
```diff
@@ -0,0 +1,45 @@
...
+                     for i in range(n_iterations)).
+The predicted class is then the argmax of the K probabilities.
+
+Regarding implementation details, the arrays
+- y_pred
+- gradients
+- hessians (for non-constant hessians)
+- raw_predictions, as returned by _raw_predict()
+are now 1D arrays of size (n_samples * n_trees_per_iteration), instead of
+just (n_samples). This doesn't change anything for binary classif or
```
ogrisel: classification

tests/test_compare_lightgbm.py
```diff
@@ -136,3 +136,67 @@ def test_same_predictions_classification(seed, min_samples_leaf, n_samples,
...
+    acc_lgbm = accuracy_score(y_train, pred_lightgbm)
+    acc_pygbm = accuracy_score(y_train, pred_pygbm)
+    np.testing.assert_almost_equal(acc_lgbm, acc_pygbm, decimal=2)
+
+    if max_leaf_nodes < 10 and n_samples >= 1000:
+
+        pred_lightgbm = est_lightgbm.predict(X_test)
+        pred_pygbm = est_pygbm.predict(X_test)
+        assert np.mean(pred_pygbm == pred_lightgbm) > .89
+
```
ogrisel: Maybe also compare the `predict_proba` outputs?

pygbm/gradient_boosting.py
NicolasHug: Well with the code as-is yes, but I think it's incorrect ^^

It should be returning classes for classification problems because the `predict_binned` output is directly used in `scorer._score_func`

```python
            if self.validation_split is not None:
                predicted_val = self._predict_binned(X_binned_val)
                score_val = scorer._score_func(y_val, predicted_val)
```

Tests don't fail because this is not tested, that's why I opened #55.

I say we leave it as-is even if it's buggy and fix everything when we address #55?

pygbm/gradient_boosting.py
```diff
@@ -285,16 +366,33 @@ def __init__(self, loss='binary_crossentropy', learning_rate=0.1,
...
     def predict(self, X):
+        # This could be done in parallel
         return np.argmax(self.predict_proba(X), axis=1)
 
     def predict_proba(self, X):
         raw_predictions = self._raw_predict(X)
         return self.loss_.predict_proba(raw_predictions)
 
+    def _get_n_trees_per_iteration(self, y):
+        n_classes = np.unique(y).shape[0]
```
ogrisel: Instead of `np.unique(y)` we should be using the `LabelEncoder` of scikit-learn to set the `classes_` attribute to allow for string class labels or non-contiguous integer class labels. Then instead of storing  of `n_trees_per_iteration_` attribute we could do:

```python
n_trees_per_iteration = 1 if self.classes_.shape[0] == 2 else self.classes_.shape[0]
```

where we actually need it.

NicolasHug: Ok I'll do that.

But we need `n_trees_per_iteration_` even in regression so I doubt we could get rid of the attribute. `GradientBoostingRegressor` has `_get_n_tree_per_iteration` that always returns `1`.

ogrisel: Ok as you prefer.

NicolasHug: I mean of course I'm open to any other suggestions. I just don't think we can easily avoid having `n_trees_per_iterations_`.

But sure I'll use `LabelEncoder` to have better support for classification

pygbm/loss.py
```diff
@@ -24,20 +40,30 @@ def _get_threads_chunks(total_size):
...
 def _expit(x):
     # custom sigmoid because we cannot use that of scipy with numba
+    # TODO make it numerically stable (if needed?)
     return 1 / (1 + np.exp(-x))
 
 
 class Loss(ABC):
 
+    def __init__(self, n_trees_per_iteration=1):
+        self.n_trees_per_iteration_ = n_trees_per_iteration
```
ogrisel: The `Loss` objective function is not an estimator, so the `n_trees_per_iteration` attribute does not have to be suffixed with an underscore. The underscore is used distinguish fitted parameters from constructor hyperparameter in scikit-learn estimators.

pygbm/loss.py
```diff
@@ -104,5 +129,67 @@ def _update_gradients_hessians_logistic(gradients, hessians, y_true, y_pred):
             hessians[i] = gradient_abs * (1. - gradient_abs)
 
 
+class CategoricalCrossEntropy(Loss):
+
+    hessian_is_constant = False
+
+    def __call__(self, y_true, y_pred, average=True):
+        y_pred = y_pred.reshape(self.n_trees_per_iteration_, -1).T
```
ogrisel: Actually, I think we should get rid off the `self.n_trees_per_iteration_` attribute completely in the loss functions. Here we could just do:

```python
 y_pred = y_pred.reshape(y_true.shape[0], -1)
```

(I think).

pygbm/loss.py
```diff
@@ -104,5 +129,67 @@ def _update_gradients_hessians_logistic(gradients, hessians, y_true, y_pred):
...
+        one_hot_true = np.zeros_like(y_pred)
+        for k in range(self.n_trees_per_iteration_):
```
ogrisel:
```python
for k in range(y_pred.shape[1]):
    ...
```

pygbm/loss.py
```diff
@@ -104,5 +125,67 @@ def _update_gradients_hessians_logistic(gradients, hessians, y_true, y_pred):
...
+    def predict_proba(self, raw_predictions, n_trees_per_iteration):
+        # TODO: This could be done in parallel
+        raw_predictions = raw_predictions.reshape(n_trees_per_iteration, -1).T
+        # compute softmax (using exp(log(softmax)))
+        return np.exp(raw_predictions -
+                      logsumexp(raw_predictions, axis=1)[:, np.newaxis])
+
+
+@njit(parallel=True)
+def _update_gradients_hessians_softmax(gradients, hessians, y_true, y_pred):
```
ogrisel: categorical

NicolasHug:
@ogrisel I renamed `y_pred` into `raw_predictions`.
It now has shape `(n_samples, n_trees_per_iteration)`.

I tried reshaping gradients and hessians as well to 2D arrays, but that made numba fail. I think it's expecting C arrays and got Fortran arrays instead (didn't looks into it in details).

I'll merge as-is, this is not blocking and we can change it later.

PR #60, opened 2018-12-07
[MRG] Pass check_estimator form scikit-learn

NicolasHug:
Most of the checks are passing, ~but still WIP.~

### DIFF
--- a/pygbm/gradient_boosting.py
+++ b/pygbm/gradient_boosting.py
@@ -97,6 +97,11 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
         # TODO: test input checking
         X, y = check_X_y(X, y, dtype=[np.float32, np.float64])
         y = self._encode_y(y)
+        if X.shape[0] == 1 or X.shape[1] == 1:
+            raise ValueError(
+                'Passing only one sample or one feature is not supported yet. '
+                'See numba issue #3569.'
+            )
         rng = check_random_state(self.random_state)
 
         self._validate_parameters()
--- a/pygbm/gradient_boosting.py
+++ b/pygbm/gradient_boosting.py
@@ -123,8 +123,6 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
             X_binned_train, X_binned_val, y_train, y_val = train_test_split(
                 X_binned, y, test_size=self.validation_split,
                 stratify=stratify, random_state=rng)
-<<<<<<< HEAD
-=======
             if X_binned_train.size == 0 or X_binned_val.size == 0:
                 raise ValueError(
                     f'Not enough data (n_samples={X_binned.shape[0]}) to '
@@ -132,7 +130,6 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
                     f'{self.validation_split}. Use more training data or '
                     f'adjust validation_split.'
                 )
->>>>>>> Fixed most checks, few more to investigate
             # Histogram computation is faster on feature-aligned data.
             X_binned_train = np.asfortranarray(X_binned_train)
         else:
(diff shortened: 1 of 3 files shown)


tests/test_gradient_boosting.py
```diff
@@ -192,6 +192,12 @@ def custom_check_estimator(Estimator):
             # X is both Fortran and C aligned and numba can't compile.
             # Opened numba issue 3569 (not sure if this is by design)
             continue
+        if check is estimator_checks.check_classifiers_train:
+            continue # probas don't exactly sum to 1 (very close though)
```
ogrisel: wouldn't it be possible to post process the softmax to renormalize instead?

NicolasHug: I guess... but that's some extra computation just for the test to pass.

The sum of probas are really close to 1:

Changing `assert np.sum(y_prob, axis=1), np.ones(n_samples))` to `assert_array_almost_equal(np.sum(y_prob, axis=1), np.ones(n_samples))` 
make the check pass.

Alternatively we could switch to float64...

pygbm/gradient_boosting.py
```diff
@@ -99,6 +100,7 @@ def fit(self, X, y):
         rng = check_random_state(self.random_state)
 
         self._validate_parameters()
+        self.n_features_ = X.shape[1]  # used for validation in predict()
```
ogrisel: Instead of introducing a new attribute in the public API I would rather add a `n_features` attribute to the `TreePredictor` class to be passed as a constructor parameter and use that to do the check in the fitted `GradientBoostingMachine` classes.

tests/test_gradient_boosting.py
```diff
@@ -171,3 +172,55 @@ def test_early_stopping_classification(data, scoring, validation_split, tol):
...
+
+    estimator = Estimator
+    name = type(estimator).__name__
+
+    for check in _yield_all_checks(name, estimator):
+        if (check is estimator_checks.check_fit2d_1feature or
+                check is estimator_checks.check_fit2d_1sample):
+            # X is both Fortran and C aligned and numba can't compile.
+            # Opened numba issue 3569 (not sure if this is by design)
+            continue
```
ogrisel: In the meantime, we could make the `GradientBoostingMachine` classes raise a `ValueError` with an informative error message.

tests/test_gradient_boosting.py
ogrisel: Alright. Maybe we could think of relaxing this check in scikit-learn when `predict_proba` has 32-bit dtype. But this is fine for now.

tests/test_gradient_boosting.py
```diff
@@ -171,3 +172,55 @@ def test_early_stopping_classification(data, scoring, validation_split, tol):
...
+        if (hasattr(check, 'func') and
+                check.func is estimator_checks.check_classifiers_train):
+            continue # same, wrapped in a functools.partial object.
+
+        try:
+            check(name, estimator)
+        except SkipTest as exception:
+            # the only SkipTest thrown currently results from not
+            # being able to import pandas.
+            warnings.warn(str(exception), SkipTestWarning)
```
ogrisel: Please add pandas to the `tox.ini` configuration so that we actually run the test with pandas. But ok to not make pandas a mandatory dependency.

pygbm/gradient_boosting.py
NicolasHug: The problem is that `raw_predict` (where we check the number of features) might be called while no `TreePredictor` has been created yet: we always compute the training / validation score of the init model, before the first boosting iteration.

ogrisel: Alright.

PR #61, opened 2018-12-09
[MRG] Added pygbm to lightgbm model conversion

NicolasHug:
Created `utils.py` with a utility that builds a lightgbm estimator with the same hyperparameters of a pygbm estimator.

### DIFF
--- a/pygbm/utils.py
+++ b/pygbm/utils.py
@@ -4,17 +4,20 @@ from lightgbm import LGBMClassifier
 from .gradient_boosting import GradientBoostingClassifier
 
 
-def get_lightgbm_estimator(pygbm_est):
-    """Return an estimator from LightGBM that has the same hyperparameters
-    as pygbm_est"""
+def get_lightgbm_estimator(pygbm_estimator):
+    """Return an unfitted LightGBM estimator with matching hyperparams.
 
-    pygbm_params = pygbm_est.get_params()
+    This utility function takes care of renaming the PyGBM parameters into
+    their LightGBM equivalent parameters.
+    """
+
+    pygbm_params = pygbm_estimator.get_params()
 
     if pygbm_params['loss'] == 'auto':
         raise ValueError('auto loss is not accepted. We need to know if '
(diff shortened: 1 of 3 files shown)


pygbm/utils.py
```diff
@@ -0,0 +1,50 @@
+from lightgbm import LGBMRegressor
+from lightgbm import LGBMClassifier
+
+from .gradient_boosting import GradientBoostingClassifier
+
+
+def get_lightgbm_estimator(pygbm_est):
+    """Return an estimator from LightGBM that has the same hyperparameters
+    as pygbm_est"""
```
ogrisel:
```python
"""Return an unfitted LightGBM estimator with matching hyperparams

This utility function takes care of renaming the PyGBM parameters into
their LightGBM equivalent parameters.

The PyGBM estimator is assumed to be unfitted (no trained trees).
"""
```

pygbm/utils.py
```diff
@@ -0,0 +1,50 @@
+from lightgbm import LGBMRegressor
+from lightgbm import LGBMClassifier
+
+from .gradient_boosting import GradientBoostingClassifier
+
+
+def get_lightgbm_estimator(pygbm_est):
```
ogrisel: `pygbm_est` => `pygbm_estimator`

pygbm/utils.py
```diff
@@ -0,0 +1,50 @@
...
+    """Return an estimator from LightGBM that has the same hyperparameters
+    as pygbm_est"""
+
+    pygbm_params = pygbm_est.get_params()
+
+    if pygbm_params['loss'] == 'auto':
+        raise ValueError('auto loss is not accepted. We need to know if '
+                         'the problem is binary or multiclass classification.')
+    if pygbm_params['scoring'] is not None:
+        raise ValueError('Early stopping should be deactivated.')
```
ogrisel: Maybe raise `NotImplementedError` instead?

NicolasHug:
Thanks for the review @ogrisel

PR #62, opened 2018-12-10
[MRG] Update bench_predictor.py and move _get_threads_chunks into utils.py

### LINKED ISSUE #48: Parallelize predictions with static scheduling
Not sure if this will have a high impact on the time, but in predictions functions we could replace the parallel loop over all samples

```python
for i in prange(numeric_data.shape[0]):
```

to a static scheduling to take advantage of the cache.

This is probably only going to be significant if `scoring is not None` because in this case `_predict_binned` will be called at every iteration.

NicolasHug:
Closes #48 

This PR updates the `bench_predictor` benchmark and moves `_get_threads_chunks` into `utils.py`.

This was initially an attempt to address #48, but it turns out that the results aren't really convincing: computation times with the static scheduling were slightly worse than the current version.

### DIFF
--- a/benchmarks/bench_predictor.py
+++ b/benchmarks/bench_predictor.py
@@ -1,46 +1,75 @@
 from time import time
+
 import numpy as np
 from numpy.testing import assert_allclose
+from sklearn.datasets import make_regression
 from pygbm.grower import TreeGrower
+from pygbm.binning import BinMapper
+from pygbm.loss import LeastSquares
 
 
 # TODO: make some random numerical data, bin it, and fit a grower with many
 # leaves in a joblib'ed function.
 
+n_samples = int(5e6)
 
-predictor = grower.make_predictor(bin_thresholds=bin_thresholds)
-binned_features_c = np.ascontiguousarray(binned_features)
+X, y = make_regression(n_samples=n_samples, n_features=5)
+bin_mapper = BinMapper()
+X_binned = bin_mapper.fit_transform(X)
+
+loss = LeastSquares()
+gradients, hessians = loss.init_gradients_and_hessians(
+    n_samples, n_trees_per_iteration=1)
+y_pred_init = np.zeros_like(y)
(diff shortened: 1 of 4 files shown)


benchmarks/bench_predictor.py
```diff
@@ -1,46 +1,75 @@
...
+    n_samples, n_trees_per_iteration=1)
+y_pred_init = np.zeros_like(y)
+loss.update_gradients_and_hessians(gradients, hessians, y, y_pred_init)
 
-predictor = grower.make_predictor(bin_thresholds=bin_thresholds)
-binned_features_c = np.ascontiguousarray(binned_features)
+grower = TreeGrower(X_binned, gradients, hessians, max_leaf_nodes=None)
+grower.grow()
+
+predictor = grower.make_predictor(bin_thresholds=bin_mapper.bin_thresholds_)
```
ogrisel: Rather than building the predictor from the low level API one could probably use the high level GradientBoostingRegressor API and then take:

```
predictor = regressor.predictors_[0]
```

benchmarks/bench_predictor.py
```diff
@@ -1,46 +1,75 @@
...
+assert_allclose(scores_binned_f, scores_binned_c)
+
+data_size = X.nbytes
+print("Computing predictions (F-contiguous numerical data)...")
+tic = time()
+scores_f = predictor.predict(np.asfortranarray(X))
 toc = time()
 duration = toc - tic
-print(f"done in {duration:.3f}s ({data_size / duration / 1e9:.3} GB/s)")
+print(f"done in {duration:.4f}s ({data_size / duration / 1e9:.3} GB/s)")
```
ogrisel: Maybe print in MB/s as we do not observe speed larger than 1GB/s on common hardware for a tree of this size and also add the speed in records/s (using `{X.shape[0] / duration:.3e} records/s`)

NicolasHug:
Thanks for the reviews!

PR #64, opened 2018-12-11
[MRG] Initialize predictions to average value or class probabilities

### LINKED ISSUE #47: Initialize predictions to something better than just 0
Saving it here so we remember: instead of initializing `y_pred` to 0 we may want to do something more sensible like boosting from the average target, or even accept an `init` predictor.

NicolasHug:
Closes #47 

Predictions are now initialized to something a bit better than 0 for faster convergence. I based the initialization off of LGBM.

LGBM was using `boost_from_average=True` by default from the beginning, even in our `test_compare_lightgbm`. Weirdly enough, now that our code is more similar to theirs, I had to relax an accuracy check from 1% to 1.1% (only for 1 seed out of the 5).

### DIFF
--- a/pygbm/gradient_boosting.py
+++ b/pygbm/gradient_boosting.py
@@ -170,7 +170,7 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
         # n_samples * n_trees_per_iteration
         gradients, hessians = self.loss_.init_gradients_and_hessians(
             n_samples=n_samples,
-            n_trees_per_iteration=self.n_trees_per_iteration_
+            prediction_dim=self.n_trees_per_iteration_
         )
 
         # predictors_ is a matrix of TreePredictor objects with shape
--- a/pygbm/gradient_boosting.py
+++ b/pygbm/gradient_boosting.py
@@ -156,15 +156,15 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
             print("Fitting gradient boosted rounds:")
 
         n_samples = X_binned_train.shape[0]
-        self.init_predictions_ = self.loss_.get_init_prediction(
+        self.baseline_prediction_ = self.loss_.get_baseline_prediction(
             y_train, self.n_trees_per_iteration_)
-        # raw_predictions are the values predicted by the trees for the
-        # training data.
+        # raw_predictions are the accumulated values predicted by the trees
+        # for the training data.
         raw_predictions = np.zeros(
             shape=(n_samples, self.n_trees_per_iteration_),
-            dtype=np.float32
+            dtype=self.baseline_prediction_.dtype
         )
-        raw_predictions += self.init_predictions_
+        raw_predictions += self.baseline_prediction_
 
         # gradients and hessians are 1D arrays of size
         # n_samples * n_trees_per_iteration
@@ -379,9 +379,9 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
         n_samples = X.shape[0]
         raw_predictions = np.zeros(
             shape=(n_samples, self.n_trees_per_iteration_),
-            dtype=np.float32
+            dtype=self.baseline_prediction_.dtype
         )
-        raw_predictions += self.init_predictions_
+        raw_predictions += self.baseline_prediction_
         # Should we parallelize this?
         for predictors_of_ith_iteration in self.predictors_:
             for k, predictor in enumerate(predictors_of_ith_iteration):
--- a/pygbm/gradient_boosting.py
+++ b/pygbm/gradient_boosting.py
@@ -142,18 +142,23 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
             print("Fitting gradient boosted rounds:")
 
         n_samples = X_binned_train.shape[0]
-        # values predicted by the trees. Used as-is in regression, and
-        # transformed into probas and / or classes for classification
+        self.init_predictions_ = self.loss_.get_init_prediction(
+            y_train, self.n_trees_per_iteration_)
+        # raw_predictions are the values predicted by the trees for the
+        # training data.
         raw_predictions = np.zeros(
             shape=(n_samples, self.n_trees_per_iteration_),
-            dtype=y_train.dtype
+            dtype=np.float32
         )
+        raw_predictions += self.init_predictions_
+
         # gradients and hessians are 1D arrays of size
         # n_samples * n_trees_per_iteration
         gradients, hessians = self.loss_.init_gradients_and_hessians(
             n_samples=n_samples,
             n_trees_per_iteration=self.n_trees_per_iteration_
         )
+
         # predictors_ is a matrix of TreePredictor objects with shape
         # (n_iter_, n_trees_per_iteration)
         self.predictors_ = predictors = []
@@ -356,6 +361,7 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
             shape=(n_samples, self.n_trees_per_iteration_),
             dtype=np.float32
         )
+        raw_predictions += self.init_predictions_
         # Should we parallelize this?
         for predictors_of_ith_iteration in self.predictors_:
             for k, predictor in enumerate(predictors_of_ith_iteration):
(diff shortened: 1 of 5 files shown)


pygbm/loss.py
```diff
@@ -109,6 +116,13 @@ def __call__(self, y_true, raw_predictions, average=True):
         loss = np.logaddexp(0, raw_predictions) - y_true * raw_predictions
         return loss.mean() if average else loss
 
+    def get_init_prediction(self, y_train, n_trees_per_iteration):
+        proba_positive_class = np.mean(y_train)
+        eps = 1e-15
```
ogrisel: Maybe use `np.finfo(y_train.dtype).eps` instead.

pygbm/loss.py
```diff
@@ -109,6 +116,13 @@ def __call__(self, y_true, raw_predictions, average=True):
...
+        proba_positive_class = np.clip(proba_positive_class, eps, 1 - eps)
+        # log(x / 1 - x) is the anti function of sigmoid
```
ogrisel: This is the link function of the Binomial predictive model.

pygbm/gradient_boosting.py
```diff
@@ -156,18 +156,23 @@ def fit(self, X, y):
             print("Fitting gradient boosted rounds:")
 
         n_samples = X_binned_train.shape[0]
-        # values predicted by the trees. Used as-is in regression, and
-        # transformed into probas and / or classes for classification
+        self.init_predictions_ = self.loss_.get_init_prediction(
```
ogrisel: I would rename this to `self.baseline_prediction_` (singular because it does not depend on the number of samples, even if for multiclass it's actually a n-dimensional single prediction :)

pygbm/loss.py
```diff
@@ -156,6 +170,19 @@ def __call__(self, y_true, raw_predictions, average=True):
         return (logsumexp(raw_predictions, axis=1) -
                 (one_hot_true * raw_predictions).sum(axis=1))
 
+    def get_init_prediction(self, y_train, n_trees_per_iteration):
+        init_value = np.zeros(
+            shape=(1, n_trees_per_iteration),
+            dtype=np.float32
+        )
+        eps = 1e-15
```
ogrisel: `np.finfo(y_train.dtype).eps`

pygbm/gradient_boosting.py
```diff
@@ -156,18 +156,23 @@ def fit(self, X, y):
...
+            y_train, self.n_trees_per_iteration_)
+        # raw_predictions are the values predicted by the trees for the
```
ogrisel: the **accumulated** values predicted by the trees...

pygbm/gradient_boosting.py
```diff
@@ -156,18 +156,23 @@ def fit(self, X, y):
...
+        # training data.
         raw_predictions = np.zeros(
             shape=(n_samples, self.n_trees_per_iteration_),
-            dtype=y_train.dtype
+            dtype=np.float32
```
ogrisel: Maybe `dtype=self.init_predictions_.dtype`?

pygbm/gradient_boosting.py
NicolasHug: So `baseline_prediction_` or `singular_prediction_`?

I think I like `baseline` better

ogrisel: baseline yes.

tests/test_loss.py
```diff
@@ -152,3 +153,32 @@ def test_logsumexp():
...
+            assert_all_finite(init)
+
+    # check that the init values are constants
+    for loss_name, n_trees_per_iteration in (
+            ('binary_crossentropy', 1), ('categorical_crossentropy', 5)):
+        loss = _LOSSES[loss_name]()
+        y_true = rng.randint(0, n_trees_per_iteration + 1,
+                             size=1000).astype(np.float32)
+        init = loss.get_baseline_prediction(y_true, n_trees_per_iteration)
+        assert np.unique(init).size == n_trees_per_iteration
```
ogrisel: Could you please split this test in 3 tests: one for each loss and actually test that the baseline prediction matches the class frequency ratios for the classification losses.

Please also add assertions on the expected shape of `loss.get_baseline_prediction(y_train, n_trees_per_iteration)` in each case.

tests/test_loss.py
```diff
@@ -152,3 +153,32 @@ def test_logsumexp():
...
+            loss = _LOSSES[loss_name]()
+            init = loss.get_baseline_prediction(y_true, n_trees_per_iteration)
+            assert_all_finite(init)
+
+    # check that the init values are constants
+    for loss_name, n_trees_per_iteration in (
+            ('binary_crossentropy', 1), ('categorical_crossentropy', 5)):
+        loss = _LOSSES[loss_name]()
+        y_true = rng.randint(0, n_trees_per_iteration + 1,
+                             size=1000).astype(np.float32)
```
ogrisel: This should be named `y_train`, right?

pygbm/loss.py
```diff
@@ -156,6 +185,19 @@ def __call__(self, y_true, raw_predictions, average=True):
         return (logsumexp(raw_predictions, axis=1) -
                 (one_hot_true * raw_predictions).sum(axis=1))
 
+    def get_baseline_prediction(self, y_train, n_trees_per_iteration):
```
ogrisel: I would rename `n_trees_per_iteration` as `prediction_dim` in the context of the loss function classes.

tests/test_loss.py
NicolasHug: >test that the baseline prediction matches the class frequency ratios for the classification losses.

You mean check that the values are exactly as expected (log(1 / 1-p))? I'm not sure how to do that without simply copy pasting the code from the loss, but then this test doesn't check much.

pygbm/loss.py
NicolasHug: I'll do it but I'm not sure it's always better. For example in `_update_gradients_hessians_categorical_crossentropy`, the update is semantically related to a tree, not to a 'dimension'.

tests/test_loss.py
ogrisel: Yes, consider this simple test as a non regression test and a way to clarify what is the intent of baseline prediction.

pygbm/loss.py
ogrisel: That's a good point. I was thinking that at some point we might want to try to implement multioutput regression trees such that we can have a categorical loss with one 1 tree per iteration. Anyways, we might rework that later.

ogrisel: My point was to rename this local variable to `prediction_dim` only in the context of the loss function but keep the `n_trees_per_iteration` in the context of gradient boosting iteration.

But I think it's also ok to use the name `raw_prediction_dim` or `prediction_dim` everywhere.

tests/test_loss.py
```diff
@@ -152,3 +153,33 @@ def test_logsumexp():
...
+    assert baseline_prediction.shape == tuple()  # float
+    assert_almost_equal(baseline_prediction, y_train.mean())
+
+
+def test_baseline_binary_crossentropy():
+    loss = _LOSSES['binary_crossentropy']()
+    for y_train in (np.zeros(shape=100), np.ones(shape=100)):
+        y_train = y_train.astype(np.float32)
+        baseline_prediction = loss.get_baseline_prediction(y_train, 1)
+        assert baseline_prediction.shape == tuple()  # float
```
ogrisel: nitpick "scalar" would be a better description than "float".

tests/test_loss.py
```diff
@@ -152,3 +153,33 @@ def test_logsumexp():
...
+
+def test_baseline_categorical_crossentropy():
+    prediction_dim = 4
+    loss = _LOSSES['categorical_crossentropy']()
+    for y_train in (np.zeros(shape=100), np.ones(shape=100)):
+        y_train = y_train.astype(np.float32)
+        baseline_prediction = loss.get_baseline_prediction(y_train,
+                                                           prediction_dim)
+        assert baseline_prediction.shape == (1, prediction_dim)
+        assert_all_finite(baseline_prediction)
```
ogrisel: For those loss test, one could check that applying `_inverse_link_function(base_line_prediction)` approximately recovers `y_train[0]`.

tests/test_loss.py
ogrisel: Or even consider this test as fine grained specification/documentation from the API consumer point of view.

ogrisel:
Thanks!

PR #73, opened 2018-12-14
make lightgbm import lazy

### LINKED ISSUE #72: All the examples require lightgbm
I tried to run the examples but they all install lightgbm which is not a requirement. Maybe make them optional?

amueller:
Fixes #72


NicolasHug:
Thanks!

PR #74, opened 2018-12-15
[MRG] Accept pre-binned data in fit

### LINKED ISSUE #68: Passing pre-binned data to `fit()`
I think that it is not possible to pass pre-binned data into `fit()`, or at least that results may not be as expected.

This is the case both in the current implementation, and also with the changes in #65.

Let's say we pre-bin `X` into `X_binned`:

```python
X_binned = initial_mapper.fit_transform(X)
```

and then call `fit(X_binned)`.

One might expect `predict(X)` to output the same as `predict_binned(X_binned)`, but that's not the case.

The reason is that in `fit()` the predictor is built with

```python
predictor = grower.make_predictor(bin_thresholds=self.bin_mapper_.bin_thresholds_)
```

and despite the idempotence of the mapper, `self.bin_mapper_.bin_thresholds_` is completely different from `initial_mapper.bin_thresholds_`.

* Do we want to raise an error if a uint8 array is passed to `fit` to avoid confusion?
* Do we want to allow passing pre-binned data and bypass the mapping in `fit()`? That would require passing `bin_thresholds` to `fit()` I think.

I'd go for the first option.

------------

I'm also thinking about adding this test to clarify the behavior. It passes on #65 branch:

```python
    X, y = make_regression()
    gbdt = GradientBoostingRegressor(scoring=None, random_state=0)
    mapper = BinMapper(random_state=0)
    X_binned = mapper.fit_transform(X)


    with pytest.raises(AssertionError):
        gbdt.fit(X_binned, y)
        np.testing.assert_array_almost_equal(gbdt.predict(X),
                                             gbdt.predict(X_binned))

    pred_X = gbdt.fit(X, y).predict(X)
    pred_X_binned = gbdt.fit(X_binned, y).predict(X_binned)
    np.testing.assert_array_almost_equal(pred_X, pred_X_binned)
```

NicolasHug:
Closes #68 

This PR makes it possible to pass pre-binned data in `fit()` based on the dtype of `X`, in which case the `BinMapper` is bypassed.

If the estimator was fitted with binned data, it raises an error when `predict` or `predict_proba` is used with real-valued data.

I renamed `bin_thresholds` into `numerical_treshold` to avoid confusion and to be consistent with the `TreePredictor` fields.

### DIFF
--- a/pygbm/gradient_boosting.py
+++ b/pygbm/gradient_boosting.py
@@ -69,7 +69,7 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
         if self.tol is not None and self.tol < 0:
             raise ValueError(f'tol={self.tol} '
                              f'must not be smaller than 0.')
-        if self.train_data_is_pre_binned_:
+        if X.dtype == np.uint8:  # pre-binned data
             max_bin_index = X.max()
             if self.max_bins < max_bin_index + 1:
                 raise ValueError(
@@ -112,11 +112,10 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
             )
         rng = check_random_state(self.random_state)
 
-        self.train_data_is_pre_binned_ = X.dtype == np.uint8
         self._validate_parameters(X)
         self.n_features_ = X.shape[1]  # used for validation in predict()
 
-        if self.train_data_is_pre_binned_:
+        if X.dtype == np.uint8:  # data is pre-binned
             if self.verbose:
                 print("X is pre-binned.")
             X_binned = X
@@ -389,7 +388,7 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
                 f'trained with {self.n_features_} features.'
             )
         is_binned = X.dtype == np.uint8
-        if not is_binned and self.train_data_is_pre_binned_:
+        if not is_binned and self.bin_mapper_ is None:
             raise ValueError(
                 'This estimator was fitted with pre-binned data and '
                 'can only predict pre-binned data as well. If your data *is* '
--- a/pygbm/gradient_boosting.py
+++ b/pygbm/gradient_boosting.py
@@ -73,8 +73,8 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
             max_bin_index = X.max()
             if self.max_bins < max_bin_index + 1:
                 raise ValueError(
-                    f'Data is pre-binned and max_bins={self.max_bins}, '
-                    f'but data has {max_bin_index + 1} bins.'
+                    f'max_bins is set to {self.max_bins} but the data is '
+                    f'pre-binned with {max_bin_index + 1} bins.'
                 )
 
     def fit(self, X, y):
--- a/pygbm/gradient_boosting.py
+++ b/pygbm/gradient_boosting.py
@@ -42,7 +42,7 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
         self.verbose = verbose
         self.random_state = random_state
 
-    def _validate_parameters(self):
+    def _validate_parameters(self, X):
         """Validate parameters passed to __init__.
 
         The parameters that are directly passed to the grower are checked in
@@ -69,6 +69,13 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
         if self.tol is not None and self.tol < 0:
             raise ValueError(f'tol={self.tol} '
                              f'must not be smaller than 0.')
+        if self.train_data_is_pre_binned_:
+            max_bin_index = X.max()
+            if self.max_bins < max_bin_index + 1:
+                raise ValueError(
+                    f'Data is pre-binned and max_bins={self.max_bins}, '
+                    f'but data has {max_bin_index + 1} bins.'
+                )
 
     def fit(self, X, y):
         """Fit the gradient boosting model.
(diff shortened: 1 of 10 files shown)


tests/test_predictor.py
```diff
@@ -44,3 +46,31 @@ def test_boston_dataset(max_bins):
...
+        numerical_thresholds=None
+    )
+
+    assert_raises_regex(
+        ValueError,
+        'This predictor does not have numerical thresholds',
+        predictor.predict, X_binned
+    )
+
+    predictor.predict_binned(X)  # No error
```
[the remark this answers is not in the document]
NicolasHug: Yes you're right, I've added the check.

And yes all the checks in `PredictorNode` are redundant since `GradientBoosting.predict()` would only call `predict_binned()` on `uint8` data and `predict()` on `non-uint8` data.

My guess is that the overhead is negligible compared to the prediction time, but no strong opinion, we can remove them and just add a comment.

pygbm/gradient_boosting.py
```diff
@@ -69,14 +69,24 @@ def _validate_parameters(self):
         if self.tol is not None and self.tol < 0:
             raise ValueError(f'tol={self.tol} '
                              f'must not be smaller than 0.')
+        if X.dtype == np.uint8:  # pre-binned data
+            max_bin_index = X.max()
+            if self.max_bins < max_bin_index + 1:
+                raise ValueError(
+                    f'Data is pre-binned and max_bins={self.max_bins}, '
+                    f'but data has {max_bin_index + 1} bins.'
```
ogrisel: In retrospect I find the phrasing confusing. May I suggest the following:

```python
                 raise ValueError(
                     f'max_bins is set to {self.max_bins} but the data is pre-binned with '
                     f' {max_bin_index + 1} bins.'
```

pygbm/grower.py
```diff
@@ -413,25 +413,30 @@ def _finalize_splittable_nodes(self):
...
+    def make_predictor(self, numerical_thresholds=None):
         """Make a TreePredictor object out of the current tree.
 
         Parameters
         ----------
-        bin_thresholds : array-like of floats, optional (default=None)
-            The actual thresholds values of each bin.
+        numerical_thresholds : array-like of floats, optional (default=None)
+            The actual thresholds values of each bin. None if the training data
+            was pre-binned.
```
ogrisel: Maybe add that the values of the array are expected to be sorted in increasing order.

pygbm/predictor.py
```diff
@@ -74,6 +79,16 @@ def predict(self, X):
...
         # (dense or sparse) on a feature by feature basis.
+
+        if not self.has_numerical_thresholds:
+            raise ValueError(
+                'This predictor does not have numerical thresholds so it can'
+                'only predict pre-binned data.'
+            )
+
+        if X.dtype == np.uint8:
+            raise ValueError('X dtype should be float32 or float64')
```
ogrisel: Maybe be more explicit:

```python
         if X.dtype == np.uint8:
             raise ValueError('X has uint8 dtype: use grower.predict_binned(X) if X is pre-binned, or'
                              ' convert X to a float32 dtype to be treated as numeral data')
```

NicolasHug:
without checks:

```
predict took 4.687ms on avg
predict_binned took 4.692ms on avg
```

with checks

```
predict took 4.934ms on avg
predict_binned took 4.860ms on avg
```

<details>

```python
from time import time

import numpy as np
from sklearn.datasets import make_regression
from pygbm import GradientBoostingRegressor
from pygbm.binning import BinMapper


max_iter = 500

X, y = make_regression(n_samples=5, n_features=100)
X_binned = BinMapper().fit_transform(X)
assert X_binned.dtype == np.uint8

gb = GradientBoostingRegressor(max_iter=max_iter, scoring=None,
                               n_iter_no_change=None)
# Compiling
gb.fit(X, y)
assert len(gb.predictors_) == max_iter
gb.predict(X)
gb.predict(X_binned)

n_exp = 100
predict_time = 0
predict_binned_time = 0
for _ in range(n_exp):

    tic = time()
    gb.predict(X)
    toc = time()
    predict_time += toc - tic
    print(f'predict took {toc - tic:.3f}s')

    tic = time()
    gb.predict(X_binned)
    toc = time()
    predict_binned_time += toc - tic
    print(f'predict_binned took {toc - tic:.3f}s')

print('-' * 10)
print(f'predict took {predict_time / n_exp * 1000:.3f}ms on avg')
print(f'predict_binned took {predict_binned_time / n_exp * 1000:.3f}ms on avg')
```

This is a very minimal overhead, right?

tox.ini
```diff
@@ -6,7 +6,7 @@ skip_missing_interpreters=True
 [testenv]
 deps =
      numpy
-     scipy
+     scipy == 1.1.0
```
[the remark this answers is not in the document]
ogrisel: Please add a comment with the url of the issue #82.

tox.ini
```diff
@@ -6,7 +6,7 @@ skip_missing_interpreters=True
 [testenv]
 deps =
      numpy
-     scipy == 1.1.0
+     scipy == 1.1.0  # temporary fix for issue #82
```
ogrisel: I think you should remove the whitespaces around "==".

tox.ini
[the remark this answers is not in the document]
NicolasHug: I needed to remove it, that was causing travis to fail :/

NicolasHug:
Thanks for the reviews!

PR #79, opened 2018-12-18
More memory friendly

maartenbreddels:
again, great project!

I'm trying to push pygbm to its limits (10^9 samples), and there are two places that can be a bit more memory friendly:
 * instead of calling choise with the full samples, and then reducing, just ask what is needed
 * the tree grower memory will only be released if the new instance is created, this causes a memory spike / some stress. Better is to enable Python to release it before the new memory is allocated. I also force gc.collect to make sure it did. (if you agree, I'll put the import toplevel)

I also that splitting.py is quite memory hungry, can this be improved? Can we not copy the gradient? It looks to me that a full copy is not needed, but you probably have more insight into this, I also wonder if partition / left_indices_buffer / right_indices_buffer can be non-full length. But i'm still digesting this piece of code.

### DIFF
--- a/pygbm/gradient_boosting.py
+++ b/pygbm/gradient_boosting.py
@@ -148,11 +148,14 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
         # Subsample the training set for score-based monitoring.
         if do_early_stopping:
             subsample_size = 10000
-            indices = np.arange(X_binned_train.shape[0])
-            if X_binned_train.shape[0] > subsample_size:
-                indices = rng.choice(indices, subsample_size)
-            X_binned_small_train = X_binned_train[indices]
-            y_small_train = y_train[indices]
+            n_samples_train = X_binned_train.shape[0]
+            if n_samples_train > subsample_size:
+                indices = rng.choice(X_binned_train.shape[0], subsample_size)
+                X_binned_small_train = X_binned_train[indices]
+                y_small_train = y_train[indices]
+            else:
+                X_binned_small_train = X_binned_train
+                y_small_train = y_train
             # Predicting is faster of C-contiguous arrays.
             X_binned_small_train = np.ascontiguousarray(X_binned_small_train)
 


pygbm/gradient_boosting.py
```diff
@@ -244,6 +242,10 @@ def fit(self, X, y):
                 _update_raw_predictions(leaves_data, raw_predictions[:, k])
                 toc_pred = time()
                 acc_prediction_time += toc_pred - tic_pred
+                # make sure we collect memory
+                del grower
```
NicolasHug: Maybe we should only collect the last grower, after the main loop?

NicolasHug:
Thanks for trying this!

Did you notice any performance change when collecting the grower, even on smaller datasets?

- partition needs to be of size `n_train_samples` all the time since this is the array that tells us to what node each sample belongs to
- the auxiliary arrays like `left_indices... right_indices...` need to be of size `n_samples_at_node`, which reduces to `n_train_samples` for the root node so we can't make them smaller.

pygbm/gradient_boosting.py
```diff
@@ -244,6 +242,10 @@ def fit(self, X, y):
...
+                import gc
+                gc.collect()
```
ogrisel: I would rather not call `gc.collect()` explicitly in a library. It can induce significant overhead if the python process has large non-collectable objects collections in memory (potentially outside of the library code).

If we want to improve the memory further here we could instead try to find a way to reuse large internal datastructures of the grower object although this would probably make the code less readable.

NicolasHug: What causes the stress is most probably the allocation of the splitter. I think it would be possible to allocate it only once in `fit()`, and pass it as an attribute to each of the `n_iter` growers.
Alternatively we could allocate only one grower, and instead of using `grower = TreeGrower(...)`, we could do `grower.reset(...)`

ogrisel: Yes I think it would make sense from a memory usage point of view.

If someone decides to work on this, it would be great to run the higgs boson benchmark script (with --no-lightgbm) under `memory_profiler` to check the impact of the refactoring.

pygbm/gradient_boosting.py
```diff
@@ -148,9 +148,7 @@ def fit(self, X, y):
         # Subsample the training set for score-based monitoring.
         if do_early_stopping:
             subsample_size = 10000
-            indices = np.arange(X_binned_train.shape[0])
-            if X_binned_train.shape[0] > subsample_size:
-                indices = rng.choice(indices, subsample_size)
+            indices = rng.choice(X_binned_train.shape[0], min(subsample_size, X_binned_train.shape[0]))
             X_binned_small_train = X_binned_train[indices]
             y_small_train = y_train[indices]
```
ogrisel: May I suggest the following:

```diff
         if do_early_stopping:
+            n_samples_train = X_binned_train.shape[0]
             subsample_size = 10000
-            indices = rng.choice(X_binned_train.shape[0], min(subsample_size, X_binned_train.shape[0]))
-            X_binned_small_train = X_binned_train[indices]
-            y_small_train = y_train[indices]
+            if n_samples_train > subsample_size:
+                indices = rng.choice(n_samples_train, subsample_size)
+                X_binned_small_train = X_binned_train[indices]
+                y_small_train = y_train[indices]
+            else:
+                X_binned_small_train = X_binned_train
+                y_small_train = y_train
```

so as to actually disable random subsampling on small datasets.

pygbm/gradient_boosting.py
maartenbreddels: Yes, gc.collect isn't great, reusing would be much nicer. It is hard for me to benchmarks, since it takes a significant amount of time to run. The difference is noticeable, I can type now, instead of the computer being seriously slow such that typing is even hard. I saw memory peaks going to 28+GB, and now it is more stable at 21 GB

pygbm/gradient_boosting.py
maartenbreddels: yes, makes more sense.

pygbm/gradient_boosting.py
ogrisel: What is the dataset size? On the Higgs boson benchmark, memory usage stays around 6GB on my laptop.

pygbm/gradient_boosting.py
maartenbreddels: Unfortunately, that would not solve it fully, the idea is:
 * I have 20GB of RAM used, the grower takes 8GB, 28GB used
 * next loop, first the new grower allocates 8GB, I have 36GB used
 * old grower goes out of scope, somewhere in the future it will be garbage collected

Even without the explicit gc.collect, the 36GB RAM situation can still occur, hope I've explained myself well.

pygbm/gradient_boosting.py
ogrisel: Or maybe you grow very large trees?

maartenbreddels: it's the iris dataset, but now repeated so it has 10^9 rows, 4 features/column, it is also using some unreleased code (vaex-ml) that does some extra dirty tricks to make pygbm play nice with vaex.

maartenbreddels: I got confused by that as well. I thought Python only does cyclic GC's routinely, and the rest is just reference counts, if 0, GC it. But without gc.collect I see mem usage peaks, and everything just behaves very nondeterministically. So I assumed for a moment that Python just collect every now and then, but it should not.

In any case, reusing the memory seems like the best solution, should I remove that commit?

ogrisel: +1

maartenbreddels:
The unittests pass locally (except lightgbm for some reason), feel free to edit the PR title to your liking.

NicolasHug:
Merging, only the loss tests were failing because of the scipy issue but it's fixed on master (I couldn't re-run travis for this PR but it should be good).

Thanks @maartenbreddels !

Issue #83, opened 2018-12-18
Optionally use left/right indices buffer


maartenbreddels:
continuing from #79

In splitting.py, the left/right_indices_buffer will use up 8GB for 10^9 rows. If that causes swapping, the performance benefit of multithreading (which requires these buffers) are most probably not worth it. Would it be an option to disable this?
Is there also a method that could without the buffer (I'm still wrapping my head around the algo, maybe you already thought about it).

ogrisel:
> In splitting.py, the left/right_indices_buffer will use up 8GB for 10^9 rows. If that causes swapping, the performance benefit of multithreading (which requires these buffers) are most probably not worth it. Would it be an option to disable this?

I don't see how we could have multithreading at that level anymore. You suggest disabling thread-based parallelism for the `split_indices` operation? Maybe that could be an option. @NicolasHug might know better how LightGBM does for this part of the code.

If we are in the 10^9 rows range, I suspect that approximate feature splits using the voting strategy described in the parallel lightgbm paper might be a better way to tackle those scales. I am not 100% sure though.

NicolasHug:
> Would it be an option to disable this?

Technically yes... I suppose we could use a single-threaded quick-sort like partitioning scheme.

>Is there also a method that could without the buffer

I don't think so, or at least no with the current strategy. those arrays are used to that sample indices don't overwrite each other

NicolasHug:
>@NicolasHug might know better how LightGBM does for this part of the code

I haven't checked again but I don't think they have an option to disable parallel splitting.

@maartenbreddels could you check if you have the same issue on LightGBM? Note that they are reusing allocated data like we plan to do in #81 so we need to take this into account

maartenbreddels:
There are some parallel in place partition algorithms: http://www.lsi.upc.edu/~lfrias/research/parpar/wea08.pdf
they don't appear super trivial, not something I'd do in 1 evening.

But I think one of the buffers could be avoided, that would already save a bit. Would you be interested in a PR that does either a single threaded split, or uses 1 buffer, or both? I can't promise I can do it, but if the 1 buffer PR makes the code less readable, and that is a reason not to merge, I won't bother.

> could you check if you have the same issue on LightGBM?

I cannot use LightGBM as it is now, the current implementation makes at least 2 memory copies, my vaex-ml hack avoids 1 copy, but still the memory usage is excessive.

[...]

PR #85, opened 2018-12-18
Single threaded split

maartenbreddels:
(continues from #79, will rebase after merge)

Implement what is discussed in #83.

TODO:
 - [ ] docstrings

### DIFF
--- a/benchmarks/bench_higgs_boson.py
+++ b/benchmarks/bench_higgs_boson.py
@@ -21,6 +21,7 @@ parser.add_argument('--no-lightgbm', action="store_true", default=False)
 parser.add_argument('--learning-rate', type=float, default=1.)
 parser.add_argument('--subsample', type=int, default=None)
 parser.add_argument('--max-bins', type=int, default=255)
+parser.add_argument('--no-parallel-split', action="store_true", default=False)
 args = parser.parse_args()
 
 HERE = os.path.dirname(__file__)
@@ -85,7 +86,8 @@ pygbm_model = GradientBoostingClassifier(loss='binary_crossentropy',
                                          max_leaf_nodes=n_leaf_nodes,
                                          n_iter_no_change=None,
                                          random_state=0,
-                                         verbose=1, parallel_splitting=False)
+                                         verbose=1,
+                                         parallel_splitting=not args.no_parallel_split)
 pygbm_model.fit(data_train, target_train)
 toc = time()
 predicted_test = pygbm_model.predict(data_test)


pygbm/splitting.py
```diff
@@ -304,6 +308,59 @@ def split_indices(context, split_info, sample_indices):
     return (sample_indices[:right_child_position],
             sample_indices[right_child_position:])
 
+@njit(parallel=False)
+def split_indices_single_thread(context, split_info, sample_indices):
+    """Split samples into left and right arrays.
+
+    This implementation requires less memory than the parallel version.
```
NicolasHug: Maybe add a comment about using Hoare's partition scheme, just for reference

pygbm/splitting.py
```diff
@@ -304,6 +308,59 @@ def split_indices(context, split_info, sample_indices):
...
+
+    Parameters
+    ----------
+    context : SplittingContext
+        The splitting context
+    split_ingo : SplitInfo
```
NicolasHug: split_info (my fault ^^)
can you also fix it at the other occurrence please?

pygbm/splitting.py
```diff
@@ -304,6 +308,59 @@ def split_indices(context, split_info, sample_indices):
...
+            j -= 1
+        if i >= j:  # j can become smaller than j!
+            break
+        else:
+            # swap
+            sample_indices[i], sample_indices[j] = sample_indices[j], sample_indices[i]
+            i += 1
+            j -= 1
+    return (sample_indices[:i],
+            sample_indices[i:])
```
NicolasHug: `return sample_indices[:i], sample_indices[i:]` ?

pygbm/grower.py
```diff
@@ -336,6 +339,7 @@ def split_next(self):
         node = heappop(self.splittable_nodes)
 
         tic = time()
+        split_indices = split_indices_parallel if self.parallel_splitting else split_indices_single_thread
```
NicolasHug: I think we should fork in `SplittingContext` instead of in the grower. We probably don't need `grower.parallel_splitting`.

That is, call `SplittingContext.split_indice()` in the grower, and in splitting.py do:


```python
def split_indice(...)
   if self.parallel_splitting:
       return ...
   else:
       return ...
````

File: tests/test_splitting.py [hunk 5]
```diff
@@ -287,6 +288,12 @@ def test_split_indices():
     assert samples_left.shape[0] == si_root.n_samples_left
     assert samples_right.shape[0] == si_root.n_samples_right
 
+    samples_left_single_thread, samples_right_single_thread = split_indices_single_thread(
```
NicolasHug: Maybe instead parametrize this test with `parallel_split=True` and `parallel_split=False`, so that both methods go through exactly the same tests (though I agree your current test also covers this implicitly)

benchmarks/bench_higgs_boson.py
```diff
@@ -85,7 +85,7 @@ def load_data():
                                          max_leaf_nodes=n_leaf_nodes,
                                          n_iter_no_change=None,
                                          random_state=0,
-                                         verbose=1)
+                                         verbose=1, parallel_splitting=False)
```
ogrisel: Maybe you can expose this a commandline flag to make it easy to profile with `mprof run benchmarks/bench_higgs_boson.py --disable-parallel-splitting` (using `memory_profiler`).

pygbm/splitting.py
```diff
@@ -304,6 +308,59 @@ def split_indices(context, split_info, sample_indices):
...
+    Returns
+    -------
+    left_indices : array of int
+        The indices of the samples in the left child. This is a view on
+        context.partition.
+    right_indices : array of int
+        The indices of the samples in the right child. This is a view on
+        context.partition.
+    """
+    X_binned = context.X_binned.T[split_info.feature_idx]
```
ogrisel: This could be named `X` directly instead.

ogrisel: Or just `binned_feature` to be more explicit and avoid the confusion with the original `X` 2D array that holds the numerical data in the public API.

NicolasHug: Just be sure you propose  `binned_feature = context.X_binned.T[split_info.feature_idx]`, right?

ogrisel: yes.

pygbm/splitting.py
maartenbreddels: Ok, I didn't know it had a name, I'm ok with adding it, but did you check that it is effectively the same algo?

pygbm/splitting.py
```diff
@@ -304,6 +312,31 @@ def split_indices(context, split_info, sample_indices):
     return (sample_indices[:right_child_position],
             sample_indices[right_child_position:])
```
ogrisel, line 314: PEP8: you need 2 blank lines to separate top level functions.

pygbm/splitting.py
NicolasHug: Yes I'm pretty sure it is https://en.wikipedia.org/wiki/Quicksort#Hoare_partition_scheme

pygbm/splitting.py
```diff
@@ -304,6 +312,31 @@ def split_indices(context, split_info, sample_indices):
...
 
+@njit(parallel=False)
```
NicolasHug, line 315: You can just use `@njit`, parallelism is off by default
```suggestion
@njit
```

pygbm/splitting.py
```diff
@@ -304,6 +312,31 @@ def split_indices(context, split_info, sample_indices):
...
+        while binned_feature[sample_indices[j]] > pivot and j >= 0:
+            j -= 1
+        if i >= j:  # j can become smaller than j!
+            break
+        else:
+            # swap
+            sample_indices[i], sample_indices[j] = sample_indices[j], sample_indices[i]
+            i += 1
+            j -= 1
+    return (sample_indices[:i], sample_indices[i:])
```
NicolasHug, line 338: no need for parenthesis

tests/test_splitting.py
```diff
@@ -261,7 +262,7 @@ def test_split_indices():
                                n_bins_per_feature,
                                all_gradients, all_hessians,
                                l2_regularization, min_hessian_to_split,
-                               min_samples_leaf, min_gain_to_split)
+                               min_samples_leaf, min_gain_to_split, True)
```
NicolasHug, line 265: Please parametrize this test with `True/False` instead and remove the checks below

pygbm/splitting.py
```diff
@@ -304,6 +312,31 @@ def split_indices(context, split_info, sample_indices):
     return (sample_indices[:right_child_position],
             sample_indices[right_child_position:])
 
+@njit(parallel=False)
+def _split_indices_single_threaded(context, split_info, sample_indices):
```
NicolasHug, line 316:
```suggestion
def _split_indices_single_threaded(context, split_info, sample_indices):
    # single-threaded partition into left and right arrays (Hoare's partition scheme)
```

PR #86, opened 2018-12-19
Add flake8 linting and code style check to travis

ogrisel:
This is done using tox.


NicolasHug:
LGTM, do we also want to have the `examples` and / or `benchmarks` folders covered by the CI?

ogrisel:
Benchmark is problematic because of the large dataset size and processing time.

Examples: why not.

ogrisel:
I have also migrated the travis ci to use github API so that we should get the "ckecks" tab working soon (hopefully).

ogrisel:
For reference I have activated ogrisel/pygbm on travis-ci.com (right now we are using travis-ci.org). There is no build yet on the new .com platform:

https://travis-ci.com/ogrisel/pygbm/

Now that pygbm is activated in both platforms, the migration from .org to .com should be automatically queued:

https://docs.travis-ci.com/user/open-source-on-travis-ci-com/#existing-open-source-repositories-on-travis-ciorg

PR #88, opened 2018-12-26
[MRG] Reuse grower and splitter memory

### LINKED ISSUE #81: Reuse grower (and thus the splitter) instead of creating a new one
A new grower (and a new SplittingContext) is created at each iteration which may cause a memory use peak on large datasets (#79).

Instead of instanciating a new grower, we could just reset it and reuse the allocated space.

See https://github.com/ogrisel/pygbm/pull/79#discussion_r242546463 to evaluate the changes

NicolasHug:
Closes #81 

Instead of creating a new grower and a new splitter for every tree, we reuse the existing objects and thus the allocated arrays (`ordered_gradients`, `partition`, etc.)

I didn't observe any significant speed improvement with:

```
benchmarks/bench_higgs_boson.py --no-lightgbm  --n-trees 100  --subsample 1000000
```

but that might come with parallelization.

I haven't run any memory usage benchmark, I'm not sure if `memory_profiler` is really suited for this.

@ogrisel if you're OK with this in general I'll fix the tests (`compare_lightgbm` is already passing) and I will parallelize `SplittingContext.reset()`.

### DIFF
--- a/benchmarks/bench_higgs_boson.py
+++ b/benchmarks/bench_higgs_boson.py
@@ -86,7 +86,10 @@ pygbm_model = GradientBoostingClassifier(loss='binary_crossentropy',
                                          n_iter_no_change=None,
                                          random_state=0,
                                          verbose=1)
-pygbm_model.fit(data_train, target_train)
+@profile
+def fit():
+    pygbm_model.fit(data_train, target_train)
+fit()
 toc = time()
 predicted_test = pygbm_model.predict(data_test)
 roc_auc = roc_auc_score(target_test, predicted_test)
--- a/pygbm/gradient_boosting.py
+++ b/pygbm/gradient_boosting.py
@@ -228,6 +228,17 @@ class BaseGradientBoostingMachine(BaseEstimator, ABC):
                                                      y_train, raw_predictions)
 
             predictors.append([])
+            grower = TreeGrower(
+                X_binned_train,
+                max_bins=self.max_bins,
+                n_bins_per_feature=n_bins_per_feature,
+                max_leaf_nodes=self.max_leaf_nodes,
+                max_depth=self.max_depth,
+                min_samples_leaf=self.min_samples_leaf,
+                l2_regularization=self.l2_regularization,
+                shrinkage=self.learning_rate,
+                hessian_is_constant=self.loss_.hessian_is_constant
+            )
 
             # Build `n_trees_per_iteration` trees.
             for k, (gradients_at_k, hessians_at_k) in enumerate(zip(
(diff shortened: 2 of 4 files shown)


benchmarks/bench_higgs_boson.py
```diff
@@ -86,7 +86,10 @@ def load_data():
                                          n_iter_no_change=None,
                                          random_state=0,
                                          verbose=1)
-pygbm_model.fit(data_train, target_train)
+@profile
```
NicolasHug, line 89: I'll remove this

ogrisel:
Here is the output of a run of Higgs boson with `memory_profiler`:

```
mprof run benchmarks/bench_higgs_boson.py --no-lightgbm --n-trees 100 --n-leaf-nodes 255
mprof plot
```

- on master:

bench_higgs_boson_master

- on this branch:

bench_higgs_boson_pr-88

So basically there is no significant difference. I think it means that the memory used by the reused splitting context instance is very little compared to the memory used by the histograms stored on each node instance. Those linked nodes in the grower tree it-self cannot be reused very easily without significant refactoring as far as I know.

ogrisel:
Also note that according to the above chart, the memory usage of pygbm is not that bad even if it's fluctuating because of the garbage collections of the temporary histogram arrays.

NicolasHug:
Thanks for the plots!

Indeed it doesn't seem to help much, at least with this number of samples.

I don't think histograms use much memory compared to the splitter though, according to my calculations:


histograms for one iteration:

- `n_bins * n_features * sizeof(HIST_DTYPE) * n_nodes = 255 * 28 * 12 * 255 = 22MB`

SplittingContext for one iteration (re-used):

-  there are 5 arrays of size uint32 or float32 (gradients, hessians, partition, left_indices_buffer, right_indices_buffer) so total =   `5 * 4 * n_samples = 220MB`

That being said, LightGBM uses an LRU cache for the histograms.

Maybe it would be helpful if @maartenbreddels  could try this branch on the big dataset that used to cause a memory stress? Until then I'm in favor for closing the PR, and maybe resuscitate it eventually if needed.

ogrisel:
Indeed you are right about the relative sizes of the datastructures. But then I don't understand what is causing the fluctuations when running the benchmark on this PR (with the reset-able splitting context variant).

NicolasHug:
Well maybe is just like you said, the `TreeNode` objects (that contain the histograms) accumulate, and are freed regularly?

maartenbreddels:
Happy to test this branch out (maybe this week? can't guarantee), note that I was using OS X, maybe that mattters.

ogrisel:
Here is a possible fix for the accumulated memory usage of the stored histograms that proved very efficient in scikit-learn: https://github.com/scikit-learn/scikit-learn/pull/18242

ogrisel:
Also note that Higgs boson is not the best benchmark for memory efficiency: it has many samples and few features and a single binary target. So compared to the amount of computation, it allocates few histograms.

